PR #2, opened 2017-10-05
Set up enzyme testing

uxengine:


components/Me/index.test.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import test from 'tape'
+import { configure, shallow } from 'enzyme'
+import Adapter from 'enzyme-adapter-react-16'
+import { Me } from '../Me'
+
+configure({ adapter: new Adapter() })
```
tpreusse: let's create a `lib/utils/enzyme.js` which does:

```
import Enzyme from 'enzyme'
import Adapter from 'enzyme-adapter-react-16'

Enzyme.configure({ adapter: new Adapter() })

export * from 'enzyme'
export default Enzyme
```

tpreusse: and then import it here

components/Me/index.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+test('utils.createActionButton', assert => {
```
tpreusse: let's add a proper name

components/Me/index.test.js
```diff
@@ -0,0 +1,25 @@
...
+  assert.plan(1)
+
+  // TODO: Properly hook up withMe/withT here.
```
tpreusse: IMO no need to test `withMe/withT` right now

tpreusse: maybe a better test case would be popover, clicking close and testing that it is hidden

tpreusse:
Reading the `next.js` 4.0.0 release notes I noticed a new features that allows to configure babel presets used by `next.js` and making them tape runner compatible:
https://github.com/zeit/next.js/pull/2991

Now we should have full next support in tests.

uxengine:
Thanks for the improvements!

PR #9, opened 2017-10-20
fix profile page for users without a testimonial

patte:


uxengine:
Thanks for the fix, Patte!

PR #11, opened 2017-10-24
Use the CommentTree components from the styleguide to render a discussion

wereHamster:
The code is split into the following files:

 - Discussion.js: renders a whole discussion, including the composer at the top, the tabs to switch between different orderings, and the tree below.
 - DiscussionCommentComposer: graphql-connected component which renders the composer at the top.
 - DiscussionTree: graphql-connected component which uses CommentTreeNode from the styleguide to render the whole tree.
 - enhancers.js: all graphql queries used by two components mentioned above are in this file, as well as all the code which manages subscriptions, fetchMore, updates to the apollo cache etc.


wereHamster:
I also created a simple index page so that users can create a new discussion and have an overview which discussions exist.

I noticed that the components don't work correctly when the user is not logged in. Will look at it in the afternoon.

lib/translations.json
```diff
@@ -303,8 +315,132 @@
       "value": "Diskussionsbeiträge"
     },
     {
-      "key": "testimonial/sequenceNumber",
-      "value": "Abo #{num}"
+      "key": "testimonial/meta/pageTitle",
```
uxengine: FYI, feel free to manually remove strings that I'm working on after running the translations script (just in case you only your changes to appear here).

PR #15, opened 2017-11-08
Show DiscussionPreferences in an Overlay

wereHamster:
The discussion preferences open an an Overlay. There is currently one checkbox to control the anonymity and a dropdown + input field to allow the user to select / edit the credential.

image

### DIFF
--- a/components/Discussion/enhancers.js
+++ b/components/Discussion/enhancers.js
@@ -31,9 +31,7 @@ query discussionMe($discussionId: ID!) {
   }
 }
 `
-export const withMe = graphql(meQuery, {
-  variables: ({discussionId}) => ({discussionId})
-})
+export const withMe = graphql(meQuery)
 
 export const commentsSubscription = gql`
 subscription discussionComments($discussionId: ID!) {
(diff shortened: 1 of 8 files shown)


components/Discussion/enhancers.js
```diff
@@ -365,3 +366,79 @@ mutation discussionSubmitComment($discussionId: ID!, $parentId: ID, $id: ID!, $c
...
+      credential {
+        description
+        verified
+      }
+    }
+  }
+}
+`
+export const withDiscussionPreferences = graphql(discussionPreferencesQuery, {
+  variables: ({discussionId}) => ({discussionId})
```
uxengine: Just an apollo newbie question:
I thought `variables` only appear within `{options: variables: {...}} `of a graphql() call, or within the data prop: `{ data: { variables } }`. Does it actually do anything here, or does it just work because graphql reads the `discussionId` from the component's ownProps anyway?

wereHamster: I'll try removing the code and see if it still works :)

wereHamster:
I was confused because apollo warns (it's actually an error) if your query declares variables which it doesn't use, but then it doesn't care that you pass additional props to the component which the query doesn't need. Hence why I tried to extract just what the query requires from the props.

PR #17, opened 2017-11-08
Use a new enhancer to inject the displayAuthor as it will be used when the current user submits a new comment

wereHamster:
The displayAuthor is currently extracted from `me` and the `discussion` rules and user preferences. But it would be better to make it a separate type with its own id so that when it's updated it changes on the whole page (also in the discussion comments themselves).

I also noticed that the string "Anonymous" is included in the server responses. Should that be translated by the server to the correct locale? Or should the client handle all localization? The string "Anonym" that is shown in the CommentComposer when the user decides to comment anonymously is pulled out of the translations (see the diff).

image

### DIFF
--- a/components/Discussion/enhancers.js
+++ b/components/Discussion/enhancers.js
@@ -55,6 +55,74 @@ query discussionMe($discussionId: ID!) {
 `
 export const withMe = graphql(meQuery)
 
+// This function extends the props with the 'DisplayUser' that will be used
+// when the current user submits a new comment in the given discussion
+// (specified by its discussionId in the ownProps). The shape of the 'DisplayUser'
+// is determined based on the discussion rules, and the users' preferences.
+//
+// The prop is called 'discussionDisplayAuthor'.
+export const withDiscussionDisplayAuthor = graphql(gql`
+query discussionDisplayAuthor($discussionId: ID!) {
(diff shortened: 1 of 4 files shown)


components/Discussion/enhancers.js
```diff
@@ -55,6 +55,74 @@ query discussionMe($discussionId: ID!) {
...
+        case 'ENFORCED': return true
+        case 'FORBIDDEN': return false
+        default: return false
+      }
+    })()
+
+    if (anonymous) {
+      return {
+        discussionDisplayAuthor: {
+          name: t('discussion/displayUser/anonymous'),
```
uxengine: Yes, I'd rely on the server response here, so you'd just have to update
https://docs.google.com/spreadsheets/d/1NE1igcbxtTrVd5P72Ubqmgv4yKYh0LRGa0EuYSEQKxE/edit#gid=0
and use that message here.

wereHamster: Ideally the whole 'DisplayUser' would be constructed by the server and sent to the client. There is a fair bit of logic here in the client that I'm sure is also implemented in the server already.

Once you extend the schema so that the client can fetch the display user, this enhancer is all you'll need to change.

PR #20, opened 2017-11-09
Discussion display author query

wereHamster:
⚠️  Requires the `displayUser-id` backend branch ⚠️

Use the `displayAuthor` field in the `Discussion` object to fetch how the comment `displayAuthor` will look if the user submits a new comment under the currently selected rules and user preferences. The `setDiscussionPreferences` mutation in turn fetches not only the new discussion preferences but also the display user object so that it is correctly updated in the local cache.

### DIFF
--- a/components/Discussion/enhancers.js
+++ b/components/Discussion/enhancers.js
@@ -63,63 +63,26 @@ export const withMe = graphql(meQuery)
 // The prop is called 'discussionDisplayAuthor'.
 export const withDiscussionDisplayAuthor = graphql(gql`
 query discussionDisplayAuthor($discussionId: ID!) {
-  me {
-    id
-    name
-    publicUser {
-      testimonial {
-        image(size: SHARE)
-      }
-    }
-  }
   discussion(id: $discussionId) {
     id
-    rules {
-      anonymity
-    }
-    userPreference {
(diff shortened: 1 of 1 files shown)


components/Discussion/enhancers.js
```diff
@@ -526,7 +501,8 @@ mutation setDiscussionPreferences($discussionId: ID!, $discussionPreferences: Di
 
           // clone() the data object so that we can mutate it in-place.
           const data = JSON.parse(JSON.stringify(immutableData))
```
tpreusse: probably a bit more efficient
```
const data = {
  ... immutableData,
  discussion: {
    ..immutableData.discussion,
    ..setDiscussionPreferences.userPreference
  }
}
```

wereHamster: I do this cloning in a few other places in the enhancer file as well. In this case your proposed change would be possible. But in the other cases (fetchMore, new comment subscription) not because I don't statically know where to modify the whole Discussion object (could be at any depth of the comments tree). So I clone the whole Discussion object, find the place where to make the change, and modify the object in-place.

tpreusse:
waiting for https://github.com/orbiting/republik-backend/pull/16

PR #23, opened 2017-11-28
Feed, a very basic one.

uxengine:
Also cleaning up utils/format.js a bit.

screen shot 2017-11-28 at 18 37 25

### DIFF
--- a/pages/feed.js
+++ b/pages/feed.js
@@ -1,0 +1,21 @@
+import React from 'react'
+import { compose } from 'redux'
+import Frame from '../components/Frame'
+import Feed from '../components/Feed'
+import Marketing from '../components/Marketing'
+import withData from '../lib/apollo/withData'
+import withMe from '../lib/apollo/withMe'
+import withT from '../lib/withT'
+
+const FeedPage = ({ url, me, t }) => {
+  const meta = {
+    title: t('pages/feed/title')
+  }
+  return (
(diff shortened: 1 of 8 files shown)


pages/feed.js
```diff
@@ -0,0 +1,21 @@
...
+import withMe from '../lib/apollo/withMe'
+import withT from '../lib/withT'
+
+const FeedPage = ({ url, me, t }) => {
+  const meta = {
+    title: t('pages/feed/title')
+  }
+  return (
+    <Frame raw url={url} meta={meta}>
+      {me && me.id ? <Feed /> : <Marketing />}
```
tpreusse: Could you port [Auth/withAuthorization](https://github.com/orbiting/publikator-frontend/blob/master/components/Auth/withAuthorization.js) and use that instead? We shouldn't make pages available under multiple urls. And a person having a deep link most likely needs to log in.

uxengine:
Done. Checking against the CF 'supporter' role for now, which may or may not change with the ongoing sell&join migration.

tpreusse:
@uxengine `supporter` gives access to the admin tools ;) but thanks for added `withAutherization`. I'll change the roles when ready

uxengine:
Of course, `supporter` is for admin tools, and `admin` is for ... whatever ;)

Removed`supporter` now.

PR #32, opened 2017-12-12
Front

tpreusse:
Front from styleguide with client side link support.


components/Front/Link.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { Link } from '../../lib/routes'
+import { parse } from 'url'
+
+import { PUBLIC_BASE_URL } from '../../lib/constants'
+import ArticleLink from '../Feed/ArticleLink'
```
uxengine: Maybe an indicator that both Feed/ArticleLink and Front/Link should live in a dedicated components/Link/ namespace.

components/Front/Link.js
```diff
@@ -0,0 +1,40 @@
...
+export default ({ href, passHref, children }) => {
+  if (!href) {
+    return children
+  }
+  const urlObject = parse(href)
+  if (urlObject.hostname && urlObject.hostname !== PUBLIC_HOSTNAME) {
+    // do nothing if url has a hostname and it's not ours
+    return children
+  }
+  // path or url is ours
```
uxengine: I'm generally OK with pattern matching external vs. internal links and logic based on that. Some of this might have to move into util methods at some point, but let's wait for the first 2nd use case to DRY things up.

PR #49, opened 2018-01-10
German paths & error page

tpreusse:
Routes diff:
https://github.com/orbiting/republik-frontend/pull/49/files#diff-52503f2920dcbf5082edfc5670ede0f1

Redirect for old ones are prepared and in our secure shared file system.

Redirect example: https://republik.love/merci

Example: https://republik.love/2017/11/28/meta

screen shot 2018-01-10 at 03 39 59


lib/routes.js
```diff
@@ -4,26 +4,26 @@ const routes = nextRoutes()
...
+  .add('media', '/medien')
+  .add('pledge', '/angebote')
+  .add('claim', '/abholen')
   .add('updates', '/updates/:slug?')
-  .add('notifications', '/notifications')
-  .add('legal/imprint', '/legal/imprint')
-  .add('legal/privacy', '/legal/privacy')
-  .add('legal/statute', '/legal/statute')
-  .add('legal/tos', '/legal/tos')
+  .add('notifications', '/notifikation')
```
clarajeanne: can we call that 'benachrichtigungen'?

components/Error/index.js
```diff
@@ -0,0 +1,89 @@
...
+const getRedirect = gql`
+query getRedirect($path: String!) {
+  redirection(path: $path) {
+    target
+    status
+  }
+}
+`
+
+const ErrorComponent = ({statusCode, t, loading, children}) => (
```
uxengine: Maybe name this something like StatusError? ErrorComponent could be all kinds of things so a more specific name could be helpful moving forward.

lib/routes.js
uxengine: or maybe just `/meldung `to keep it short?

tpreusse: I like to word but consider it too newsy to use here—«eine Republik Meldung» should just be a sign in confirmation^^

uxengine: maybe /mitteilung then, like the page title?

tpreusse: mitteilung it is

components/Error/Frame.js
```diff
@@ -0,0 +1,157 @@
...
+            {children}
+          </NarrowContainer>
+        </div>
+      </Fragment>
+    )
+  }
+}
+
+const y = (scroll, pos) => {
+  return (pos < scroll ? pos : scroll) * -0.55
```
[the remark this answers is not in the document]
tpreusse: Agree but going to skip it for now.

PR #51, opened 2018-01-10
 Display current member count on marketing page

uxengine:
Depends on https://github.com/orbiting/republik-backend/pull/45
screen shot 2018-01-10 at 14 51 14
(the count is off because it's my local DB snapshot from the past)

Also fixed missing bottom padding:
screen shot 2018-01-10 at 15 44 51


components/Marketing/index.js
```diff
@@ -80,53 +88,68 @@ const styles = {
...
-      <div {...styles.preview}>
-        <Interaction.H3 style={{ marginBottom: '17px' }}>
-          {t('marketing/preview/title')}
-        </Interaction.H3>
-        <PreviewForm />
+      <Loader error={data.error} loading={data.loading} style={{minHeight: 200}} render={() => (
+        <P {...styles.text}>
+          <RawHtml
+            dangerouslySetInnerHTML={{
+              __html: t('marketing/intro', {count: data.statistics.memberCount})
```
tpreusse: format the number here, use `countFormat` from `lib/utils/format`

components/Marketing/index.js
```diff
@@ -165,4 +170,15 @@ const MarketingPage = ({ t, crowdfundingName }) => (
...
+query membershipStats {
+  membershipStats {
+    count
+  }
+}
+`
+
+export default compose(
+  withT,
+  graphql(query)
```
tpreusse: lib constants
```
export const STATUS_POLL_INTERVAL_MS = +ENV.STATUS_POLL_INTERVAL_MS || 0
```

import `STATUS_POLL_INTERVAL_MS` here and do
```
  graphql(query, {
      options: {
        pollInterval: STATUS_POLL_INTERVAL_MS
      }
  })
```

PR #59, opened 2018-01-12
Update email

lukasbuenger:
Adds a new option to change your email address to the Account section. You immediately will get logged out afterwards.


lukasbuenger:
image

Isolated

components/Account/UpdateMe.js
```diff
@@ -150,10 +151,10 @@ class UpdateMe extends Component {
     const errorMessages = Object.keys(errors)
       .map(key => errors[key])
       .filter(Boolean)
-
     return (
       <Loader loading={loading || !me} error={error} render={() => (
         <div style={style}>
+          <UpdateEmail t={t} me={me} />
```
tpreusse: nit: you already have `t` from `withT`, use `withMe` in `UpdateEmail` and make it completely prop less.

lukasbuenger: Will that not trigger an unnecessary query?

tpreusse: apollo does query de-deup—will just load it from the in-memory cache

lukasbuenger: And I'm not sure whether "sign out via refetch" still would work.

PR #63, opened 2018-01-14
Small login form ux improvement

pozylon:
Using form onSubmit is much better:

- for supporting visible impaired people
- allows for using the return key and the submit button on native touch keyboards


pozylon:
i did but please test it too :)

tpreusse:
@pozylon I did—but with a tired mind ;)

PR #66, opened 2018-01-23
thick color line for format articles and overview

uxengine:
screen shot 2018-01-23 at 11 34 11
screen shot 2018-01-23 at 11 34 14
screen shot 2018-01-23 at 11 34 58
screen shot 2018-01-23 at 11 35 39

### DIFF
--- a/components/Article/Page.js
+++ b/components/Article/Page.js
@@ -103,6 +103,7 @@ const getDocument = gql`
             discussionId
           }
         }
+        color
         format {
           meta {
             path
@@ -250,9 +251,12 @@ class ArticlePage extends Component {
       />
     ) : null
 
-    const formatColor =
-      (meta && meta.format && meta.format.meta && meta.format.meta.color) ||
-      (article.content && article.content.meta && article.content.meta.color)
+    const formatMeta = meta && (
+      meta.template === 'format'
+        ? meta
+        : meta.format && meta.format.meta
+    )
+    const formatColor = formatMeta && formatMeta.color
 
     return (
       <Frame
--- a/components/Article/Page.js
+++ b/components/Article/Page.js
@@ -250,6 +250,10 @@ class ArticlePage extends Component {
       />
     ) : null
 
+    const formatColor =
+      (meta && meta.format && meta.format.meta && meta.format.meta.color) ||
+      (article.content && article.content.meta && article.content.meta.color)
+
     return (
       <Frame
         raw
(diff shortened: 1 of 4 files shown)


components/Frame/Header.js
```diff
@@ -36,11 +36,11 @@ const styles = {
   }),
   barOpaque: css({
     backgroundColor: '#fff',
+    boxSizing: 'content-box',
     height: HEADER_HEIGHT_MOBILE,
```
tpreusse: subtract one px border here

components/Frame/Header.js
```diff
@@ -36,11 +36,11 @@ const styles = {
...
     [mediaQueries.mUp]: {
       height: HEADER_HEIGHT
```
tpreusse: subtract one px border here

components/Frame/Popover/index.js
```diff
@@ -19,14 +19,14 @@ const menuStyle = css({
   },
   display: 'flex',
   boxSizing: 'border-box',
-  top: HEADER_HEIGHT_MOBILE,
+  top: HEADER_HEIGHT_MOBILE + 1,
```
tpreusse: let's keep the const, used elsewhere too, 100% accurate. See above

components/Frame/Popover/index.js
```diff
@@ -19,14 +19,14 @@ const menuStyle = css({
...
   left: 0,
   height: `calc(100vh - ${HEADER_HEIGHT_MOBILE}px)`,
   width: '100vw',
   flexDirection: 'column',
   padding: 20,
   [mediaQueries.mUp]: {
-    top: HEADER_HEIGHT,
+    top: HEADER_HEIGHT + 1,
```
tpreusse: rm

components/Article/Page.js
```diff
@@ -250,6 +250,10 @@ class ArticlePage extends Component {
       />
     ) : null
 
+    const formatColor =
```
tpreusse: Something like this would be more explicit and avoid `article.content.meta` (an internal, you should add color to the main meta query):
```js
const formatMeta = meta && (
  meta.template === 'format'
    ? meta
    : meta.format && meta.format.meta
)
const formatColor = formatMeta && formatMeta.color
```

uxengine:
As an alternative, I could also decouple the border from the header bar completely and render a separate DOM element for the 1px/3px border, positioned 70/45px top and "extending" downwards in case of 3px.

PR #71, opened 2018-02-05
Adapt merci title+lead depending on whether user becomes a member

uxengine:
Merci with membership:
screen shot 2018-02-05 at 18 05 09

Merci without membership:
screen shot 2018-02-05 at 18 03 30

### DIFF
--- a/components/Pledge/Form.js
+++ b/components/Pledge/Form.js
@@ -169,7 +169,6 @@ class Pledge extends Component {
               pkg => pkg.name === query.package.toUpperCase()
             )
           : null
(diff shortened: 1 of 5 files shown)


uxengine:
I've removed the query parameter now and query the pledge by ID.

tpreusse:
One more thing: Isolate the query to a `MerciText` component that is only rendered (and therefore the query only runs) when in the last case of `Merci`—actually showing the thank you text. Currently `/konto?claim=${pledgeId}` probably doesn't work anymore.

PR #73, opened 2018-02-06
new area link for front

tpreusse:
make author and format link properly clickable

### DIFF
--- a/components/Link/Area.js
+++ b/components/Link/Area.js
@@ -1,0 +1,75 @@
+import React, { Component, Children } from 'react'
+import PropTypes from 'prop-types'
+import { Router } from '../../lib/routes'
+
+const hasAncestor = (node, predicate) => {
+  if (predicate(node)) {
+    return true
+  }
+  if (node.parentNode) {
+    return hasAncestor(node.parentNode, predicate)
+  }
+  return false
+}
+
+class AreaLink extends Component {
+  constructor (props, ...rest) {
+    super(props, ...rest)
+    this.linkClicked = this.linkClicked.bind(this)
+  }
+  linkClicked (e) {
+    if (hasAncestor(e.target, node => node.nodeName === 'A')) {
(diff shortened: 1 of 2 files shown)


components/Link/Area.js
```diff
@@ -0,0 +1,75 @@
...
+  return false
+}
+
+class AreaLink extends Component {
+  constructor (props, ...rest) {
+    super(props, ...rest)
+    this.linkClicked = this.linkClicked.bind(this)
+  }
+  linkClicked (e) {
+    if (hasAncestor(e.target, node => node.nodeName === 'A')) {
```
uxengine: This is the main difference I see compared to https://github.com/zeit/next.js/blob/85f341c9d834114be1127cf73a04bc3517cc2f67/lib/link.js

Would be cool to get the optional early return for link ancestors into next.js at some point :)

tpreusse: It's hard to know what's right here. The issue comes from nesting `<Link>` elements.
```js
<Link href='a'>
<teaser>
  <Link href='b'><a>Format</a></Link>
  <Link href='a'><a>Title</a></Link>
</teaser>
</Link>
```
`A` elements can't be nested at all (by html spec). If Next.js `<Link>` elements should ever be nested is not specified.

The easiest fix would be to add a `e.stopPropagation()` here: 
https://github.com/zeit/next.js/blob/7a1d9c18a39860f5711446595543468bfbdef04e/lib/link.js#L46

But someone may want/depend on the propagated click event (although the use case is not clear to me). A generic `onClick` handler in the Next.js `<Link>` (called first) could allow for custom behaviour. But ultimately shipping a completely custom `<Link>` everywhere eventually could have some benefits too: prefetching data on hover or mouse down and direkt `next-routes` support (`next-routes` [decorades the link](https://github.com/fridays/next-routes/blob/master/src/index.js#L90) as well).

PR #74, opened 2018-02-08
add formats overview page

uxengine:
Text copy still in progress ...

FYI, I forced meta headline style on editorial format headlines, since it looks inconsistent to render a format name in title serif while it's usually sans-serif. We could also just use the normal format style, but that looks to small IMHO when the focus is on the formats themselves.

Up for discussion: Show or hide dates? We didn't want them originally but now that we sort by date ...

screen shot 2018-02-08 at 15 19 30
screen shot 2018-02-08 at 15 19 37
screen shot 2018-02-08 at 15 20 18

### DIFF
--- a/components/Formats/index.js
+++ b/components/Formats/index.js
@@ -95,7 +95,7 @@ class Formats extends Component {
                     {t('formats/title/editorial')}
                   </Interaction.H2>
                   {editorialNodes.map(doc => (
-                    <TeaserFeed {...doc.meta} Link={Link} key={doc.meta.path} />
+                    <TeaserFeed {...doc.meta} publishDate={null} Link={Link} key={doc.meta.path} />
                   ))}
                 </Fragment>
               )}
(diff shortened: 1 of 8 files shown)


components/Formats/index.js
```diff
@@ -0,0 +1,119 @@
...
+                  __html: t('formats/lead')
+                }}
+              />
+              {!!editorialNodes.length && (
+                <Fragment>
+                  <Interaction.H2 {...styles.h2}>
+                    {t('formats/title/editorial')}
+                  </Interaction.H2>
+                  {editorialNodes.map(doc => (
+                    <TeaserFeed {...{...doc.meta, kind: 'meta'}} Link={Link} key={doc.meta.path} />
```
tpreusse: `<TeaserFeed {...doc.meta} kind='meta'` would be cleaner.

However we should also be consistent with this behaviour (have the same in the main feed).

Possible solution:
- Add a template prop to `TeaserFeed`
- expend the [interaction headline clause](https://github.com/orbiting/styleguide/blob/11a30d52a2107ff2a127983c2c4cfea4a2fbc510/src/components/TeaserFeed/index.js#L53) with `template === 'format'`
- query template in `pages/feed`, hard code template='format' here or also query above

uxengine:
Format page teasers are consistent between formats overview and feed page now:
(Normal article on top, format page below)
screen shot 2018-02-08 at 16 42 49

uxengine:
@tpreusse While debating whether or not to show the date on Format teasers on the feed, we decided *not* to show them at all in the main feed for now (without date they would appear even more random). So basically we consider the formats overview page the feed for formats.

PR #76, opened 2018-02-16
add new team members and support deeplinking to /crew#name-slug

uxengine:
- Some descriptions may be subject to change by Monday (e.g. Robin's not ready yet).
- Will allow to deeplink to e.g. republik.ch/crew#christof-moser resulting in:
screen shot 2018-02-16 at 16 58 09
screen shot 2018-02-16 at 16 57 54


lib/utils/slugify.js
```diff
@@ -0,0 +1,9 @@
+import slugify from 'slugify'
```
tpreusse: please use this function instead: https://github.com/orbiting/publikator-frontend/blob/master/lib/utils/slug.js (it's even tested)

(move to styleguide or copy & paste)

components/Portrait.js
```diff
@@ -35,7 +45,7 @@ const portraitImageRightStyle = css({
 })
 
 const Portrait = ({odd, image, description, name, age, title, email}) => (
-  <div {...portraitStyle}>
+  <div {...portraitStyle} id={slugify(name)}>
```
tpreusse: normally I'd use a relative positioned anchor tag (`top: -headerHeight`):
https://github.com/orbiting/republik-frontend/blob/f190f5208d2fc28d6f07e927c85b9c4d854cb36c/components/Faq/List.js#L127
but if a `paddingTop` value of exactly the header height is what you want I'm also fine with the current solution

lib/utils/slugify.js
tpreusse: keeping dependencies at bay ;)

PR #82, opened 2018-03-05
disable ssr caching as it ignores accept headers

lukasbuenger:


components/Article/Page.js
```diff
@@ -321,12 +320,10 @@ class ArticlePage extends Component {
           return (
             <Fragment>
               {!isFormat && <PayNote.Before />}
-              <SSRCachingBoundary cacheKey={article.id}>
-                {() => renderMdast({
-                  ...article.content,
-                  format: meta.format
-                }, schema)}
-              </SSRCachingBoundary>
+              {() => renderMdast({
```
tpreusse: `SSRCachingBoundary` takes a function as a child for lazy computing (only on cache miss). When removing `SSRCachingBoundary` you need to also remove the anonymous function.

screen shot 2018-03-12 at 04 05 24

tpreusse:
Please add webp to the cache key instead: https://github.com/orbiting/republik-frontend/pull/85

PR #90, opened 2018-03-28
Paged and Sectioned Feed

ruggedly:
Add cursor & section headlines (date).

ToDo:
- [x] auto load next page on scroll (infinite scroll)
  - [x] stop after ~100, ask before loading more
- [x] sticky dates (iOS table headings like)


garamond:
infscroll

components/Feed/index.js
```diff
@@ -1,34 +1,63 @@
...
+import StickyHeader from './StickyHeader'
+import PropTypes from 'prop-types'
 
 import {
   Center,
   TeaserFeed,
   Interaction,
   mediaQueries
 } from '@project-r/styleguide'
+import Button from '@project-r/styleguide/lib/components/Button'
```
uxengine: I think you can just add `Button` to the styleguide import statement above.

components/Feed/index.js
```diff
@@ -116,20 +178,37 @@ class Feed extends Component {
...
-                    kind={
-                      doc.meta.template === 'editorialNewsletter' ? (
-                        'meta'
-                      ) : (
-                        doc.meta.kind
+                groupByDate.entries(nodes).map(({key, values}) =>
+                  <section>
+                    <div {...styles.header}>
+                      <StickyHeader>
+                        <span {...styles.date}>
```
uxengine: `styles.date` doesn't exist, so you can remove the whole `<span>` I guess.

components/Feed/index.js
```diff
@@ -61,15 +90,51 @@ const greetingSubscription = gql`
...
+    }
+    this.onScroll = () => {
+      const { infiniteScroll } = this.state
+      const { loadMore, hasMore, maxPages } = this.props
+      const d = document.documentElement
+      const scrollTop = Math.max(window.pageYOffset, d.scrollTop, document.body.scrollTop)
+      const offset = scrollTop + window.innerHeight
+      const height = d.offsetHeight
+      const needsMore = height - offset < (height / 3)
+      if ((infiniteScroll || this.pagesLoaded < maxPages) && hasMore && needsMore) {
```
uxengine: If I understand correctly, `maxPages` is only relevant until the user has activated infinite scroll. I was slightly confused in the beginning about how infinity and maxPages relate to each other. So maybe we could use a slightly more specific name like `maxInitialPages` to make that clearer at first read.

lib/translations.json
```diff
@@ -2174,6 +2174,10 @@
       "key": "format/feed/payNote",
       "value": "Sie brauchen eine Mitgliedschaft, um die Beiträge zu lesen."
     },
+    {
+      "key": "format/feed/loadMore",
+      "value": "Feed fortsetzen"
```
uxengine: Sounds a little too technical IMHO, how about something like "Mehr laden" or just "Mehr"?

components/Feed/index.js
```diff
@@ -138,4 +217,47 @@ class Feed extends Component {
...
+      props: ({ data }) => ({
+        data,
+        hasMore: data.documents && data.documents.pageInfo.hasNextPage,
+        loadMore: () => {
+          return data.fetchMore({
+            updateQuery: (previousResult, { fetchMoreResult }) => {
+              const nodes = [
+                ...previousResult.documents.nodes,
+                ...fetchMoreResult.documents.nodes
+              ].filter((node, index, all) => all.findIndex(n => n.id === node.id) === index) // deduplicating due to off by one in pagination API
```
uxengine: Can you break into one or two extra lines for readability?

components/Feed/StickyHeader.js
```diff
@@ -0,0 +1,79 @@
+import { Component } from 'react'
+import { HEADER_HEIGHT, HEADER_HEIGHT_MOBILE } from '../constants'
+import { css } from 'glamor'
+import { mediaQueries } from '@project-r/styleguide/lib/lib'
```
uxengine: `import { mediaQueries } from '@project-r/styleguide'`

components/Feed/StickyHeader.js
```diff
@@ -0,0 +1,79 @@
...
+
+const stickyHeaderStyle = {
```
uxengine: We typically name the default styles object just `styles` to avoid lengthy references like`stickyHeaderStyle.sticky`.

components/Feed/StickyHeader.js
```diff
@@ -0,0 +1,79 @@
...
+  base: css({
+    paddingTop: '8px',
+    paddingBottom: '8px',
```
uxengine: Optional:
use shorthand: `padding: '8px 0'`

components/Feed/index.js
garamond: 👍

components/Feed/index.js
garamond: 👌

components/Feed/index.js
garamond: Good point

lib/translations.json
garamond: 💯

garamond:
scrollmobile

uxengine:
Thanks, LGTM!

components/Feed/index.js
```diff
@@ -138,4 +243,50 @@ class Feed extends Component {
   }
 }
 
-export default compose(graphql(getDocuments))(Feed)
+Feed.propTypes = {
+  data: PropTypes.object.isRequired,
+  loadMore: PropTypes.func.isRequired,
+  hasMore: PropTypes.bool,
+  maxInitialDocs: PropTypes.number,
```
tpreusse: unused? rm

components/Feed/index.js
```diff
@@ -61,15 +92,63 @@ const greetingSubscription = gql`
...
+    }
+    this.isAutoLoadEnabled = () => {
+      const { infiniteScroll } = this.state
+      const { maxInitialDocs } = this.props
+      return (infiniteScroll || this.nodes.length < maxInitialDocs)
+    }
+    this.getRemainingDocumentsCount = (nodes) => {
+      const { data: { documents } } = this.props
+      return (documents.totalCount) - // all docs
+              this.nodes.length - // already displayed
```
tpreusse: use `nodes` from args

components/Feed/index.js
```diff
@@ -61,15 +92,63 @@ const greetingSubscription = gql`
...
 `
 
+const dateFormat = timeFormat('%A,\n%d.%m.%Y')
+
+const groupByDate = nest().key(d => dateFormat(new Date(d.meta.publishDate)))
+
 class Feed extends Component {
+  constructor (props) {
+    super(props)
+    this.nodes = []
```
tpreusse: I think this shouldn't be an instance variable. If you wanted to have access to the filtered nodes array: move the filter to the `props` fn of the `graphql` wrapper:
```js
      props: ({ data }) => ({
        data,
        nodes: data.documents && data.documents.nodes.filter(), // is safe because `DocumentConnection.nodes` defined as a mandatory array (`[]!`)
```

components/Feed/index.js
```diff
@@ -1,34 +1,65 @@
...
 
 const getDocuments = gql`
-  query getDocuments {
+  query getDocuments($cursor: String) {
     greeting {
       text
       id
     }
-    documents(feed: true) {
+    documents(feed: true, first: 10, after: $cursor) {
```
tpreusse: should be 50?

components/Feed/StickyHeader.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  render () {
+    const { children } = this.props
+    const { sticky } = this.state
+    return (
+      <div
+        ref={this.setRef}
+        {...style.base}
+        className={sticky ? `${style.sticky}` : ''}
```
tpreusse: This does not guarantee that `position: 'fixed'` wins. It's does happen because the base `css({...})` call is before the sticky `css({...})` call.

The situation as an static example:
https://codepen.io/anon/pen/GGEjrG

Two options to ensure it:
```js
{...merge(style.base, sticky && style.sticky)}
```
_Import merge from glamor. This comes at a slight runtime cost (hashing the css and injecting it the first time)._

Remove position from the glamor css and handle it with `style`:
```ks
{...style.base}
{...(sticky && style.sticky)}
style={{position: sticky ? 'fixed' : 'relative'}}
```
_This is faster at runtime, afaik._

Spreading `false`, `undefined` and `null` is safe:
```js
<div {...(false)} />
<div {...(undefined)} />
<div {...(null)} />
```
_Does nothing because object spread behaves like Object.assign_

garamond: Let's handle it with the `style` prop – makes CSS loading efficient and the effect of the `sticky` flag explicit.

uxengine:
iPhone SE should be OK now:
screen shot 2018-06-18 at 14 14 50

PR #91, opened 2018-04-09
discussion: maxLength, etiquette link, default sort hottest

uxengine:
Depends on https://github.com/orbiting/styleguide/pull/126

screen shot 2018-04-10 at 16 58 00

screen shot 2018-04-10 at 16 55 53

### DIFF
--- a/components/Article/Page.js
+++ b/components/Article/Page.js
@@ -15,6 +15,7 @@ import DiscussionIconLink from '../Discussion/IconLink'
 import Feed from '../Feed/Format'
 import StatusError from '../StatusError'
 import SSRCachingBoundary, { webpCacheKey } from '../SSRCachingBoundary'
+import withMembership from '../Auth/withMembership'
 
 import {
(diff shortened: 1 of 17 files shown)


tpreusse:
One more thing, if you have spare time: We forgot about another limit: `discussion.rules.minInterval`. Allowing to e.g. limit a discussion to one comment per user per day. There is a `discussion.userWaitUntil` date in the API. We should replace the comment composer with a text saying «Sie können erst in {x} Minuten/Stunden/Tagen wieder einen Beitrag schreiben.» while `discussion.userWaitUntil` is not null and in the future. The message should probably only be shown in place of the main/root composer and reply buttons grayed out with a title with the same message.

uxengine:
@tpreusse userWaitUntil is implemented now:
screen shot 2018-04-11 at 16 59 31

I'd suggest to roll with this before my time-off – mdast work hasn't even started on the frontend yet.

PR #92, opened 2018-04-22
render extract from document

tpreusse:
This allow to render extracts from a article without any frame.

Example URL:
https://republik.love/2018/04/16/pdf-fixture?extract=3:16
As an Image:
via CDN

Ranges are also supported: https://republik.love/2018/04/16/pdf-fixture?extract=3:14...21

screen shot 2018-04-23 at 01 22 22

Will be used by the PDF-Renderer for HTML-Illustrations and other elements without an PDF component.

### DIFF
--- a/components/Article/Extract.js
+++ b/components/Article/Extract.js
@@ -1,0 +1,46 @@
+import React, { Fragment } from 'react'
+import Head from 'next/head'
+import { renderMdast } from 'mdast-react-render'
+
+const parseRanges = ranges => (
+  ranges.split(':').map(range => {
+    let [start, end] = range.split('...')
+    start = +start || 0
+
+    if (end === '') {
+      end = undefined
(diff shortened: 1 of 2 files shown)


components/Article/Extract.js
```diff
@@ -0,0 +1,46 @@
...
+import Head from 'next/head'
+import { renderMdast } from 'mdast-react-render'
+
+const parseRanges = ranges => (
+  ranges.split(':').map(range => {
+    let [start, end] = range.split('...')
+    start = +start || 0
+
+    if (end === '') {
+      end = undefined
```
uxengine: This logic might is a bit more obvious if it's clear the values end up in Array.slice(), so parseSlices() would sound clearer to me.

tpreusse: improved naming and added tests

PR #101, opened 2018-05-07
fix: Accept base64-encoded email in query param, and render properly

patrickvenetz:
Ensures a passed string is decoded if it recognizes string as valid base64 chars in `SignIn` component.

A future version of backend will provide a URL with `email=<base64-encoded email>` instead of `email=<email>`. Latter let to double-encoding conflicts in some email clients, and resulted in URLs being invalid:

`heidi@wiese.tld` is encoded to `heidi%40wiese.tld` (`@` => `%40`). Some email clients would re-encode and generate a link with `heidi%2540wiese.tld` (`%` => `%25`).

Introduces [urlsafe-base64](https://www.npmjs.com/package/urlsafe-base64) as a dependency.

### DIFF
--- a/components/Auth/SignIn.js
+++ b/components/Auth/SignIn.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import { graphql, compose } from 'react-apollo'
 import gql from 'graphql-tag'
 import {css} from 'glamor'
+import urlsafeBas64 from 'urlsafe-base64'
 import { Router } from '../../lib/routes'
 import withT from '../../lib/withT'
 import isEmail from 'validator/lib/isEmail'
@@ -52,12 +53,19 @@ class SignIn extends Component {
   constructor (props) {
     super(props)
     this.state = {
-      email: props.email || '',
+      email: this.ensureDecodedEmail(props.email) || '',
       polling: false,
       loading: false,
       success: undefined
     }
   }
+
+  ensureDecodedEmail (email) {
+    return email && urlsafeBas64.validate(email)
+      ? urlsafeBas64.decode(Buffer.from(email)).toString()
+      : email
+  }
+
   render () {
     const {t, label} = this.props
     const {
--- a/components/Auth/SignIn.js
+++ b/components/Auth/SignIn.js
@@ -49,19 +49,18 @@ const styles = {
   })
 }
 
+const ensureDecodedEmail =
+  (email = '') => match(email) ? decode(email) : email
+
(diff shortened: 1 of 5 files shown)


tpreusse:
Code in the front-end need to be browser compatible, `urlsafe-base64` is not (Buffer is a node.js only global).

Luckily there window.btoa & atob. Use something like this (untested, add to utils as `base64u.js`, not a huge fan of micro package):
```js
const urlDecode = base64u => `${base64u}${'==='.slice((base64u.length + 3) % 4)}`
  .replace(/-/g, '+')
  .replace(/_/g, '/')

const urlEncode = string => string
  .replace(/\+/g, '-')
  .replace(/\//g, '_')
  .replace(/=/g, '')

export const encode = string => urlEncode(typeof window
  ? window.btoa(string)
  : Buffer.from(string, 'utf8').toString('base64')
)

export const decode = base64u => typeof window
  ? window.atob(urlDecode(base64u))
  : Buffer.from(urlDecode(base64u), 'base64').toString('utf8')

export const match = base64u => /^[A-Za-z0-9\-_]+$/.test(base64u)
```

patrickvenetz:
Thx @tpreusse. Will adopt.

patrickvenetz:
@tpreusse Before I adopt: Could I presuade you to have this package included? https://github.com/feross/buffer It is expectionally well maintained, Feross is exceptionally smart, it would prevent a "micro package". Down side is that larger footprint.

tpreusse:
[feross/buffer](https://github.com/feross/buffer) looks very nice and well maintained. But we don't need it—therefore it's 5.7kb too big. My dislike of micro packages doesn't stem from them being too small but from the security risk of using packages maintained by lone wolves for 1 to 20 lines of simple code ;)

components/Auth/SignIn.js
```diff
@@ -52,12 +53,17 @@ class SignIn extends Component {
...
     this.state = {
-      email: props.email || '',
+      email: this.ensureDecodedEmail(props.email) || '',
       polling: false,
       loading: false,
       success: undefined
     }
   }
+
+  ensureDecodedEmail (email) {
```
tpreusse: this function does not need to be on the class, you could do a small helper before the class:
```js
const ensureDecodedEmail = (email = '') => match(email) ? decode(email) : email
```
_the default argument value of `''` also makes the undefined checks and default via `or` unnecessary_

patrickvenetz:
I've made fire

PR #103, opened 2018-05-09
feat: search and results UI

uxengine:
Depends on https://github.com/orbiting/backends/pull/19

Major components:
- index.js: Manages state of user search, passes down callback functions for updating state based on user interactions and renders Input and Results components
- Results.js: Fires two kinds of graphql queries (aggregations only while typing, aggregations and results on submit) and renders Filter, Sort and Teaser components.

### Filter updates while typing:
screen shot 2018-06-12 at 10 54 19

### After submit:
screen shot 2018-06-12 at 10 54 33

### Filter selected:
screen shot 2018-06-12 at 10 54 57

## TODO
- [x] Reset button input field
- [x] Functional bucket filters
- [x] Functional sort
- [x] Submit search input on return
- [x] Navigation entry point in header according to design
- [x] Audio filter
- [x] Make search UI closable (overload X when input is empty)
- [x] Paginated results
- [x] exclude front results with negative filter
- [x] Use URL parameters/fragments to create bookmarkable searches
- [x] Show aggregations for current results ("drill-down"), don't preserve search query aggregates
- [x] debounced filter refresh
- [x] hide sort when disabled (instead of just disabling)
- [x] Open design questions (comment teaser structure, highlighting, empty results)
- [x] Highlighting in article and comment teasers
- [x] Hide feed when search form is visible. It may lead to confusion
- [x] Drop date in format teasers. Date is not self-explanatory in what it refers to 
- [x] Use new comment teaser text without markdown once backend supports it
- [x] Filter view on mobile is covering too much of screen. Single, scrollable lines of filter is recommended, UI mockup suggest a "more" button
- [x] Clicking "Load more" at end of page makes the page jump to top because of the new input focus
- [x] /suche instead of /feed?search=
- [x] manage input focus (with special behavior on mobile)
- [x] bug: search for sth, click on result, go back: "relevance" sort is not preserved
- [x] Reset filters when search query is refined
- [x] stabilize results position while typing
- [x] basic a11y check and fixes
- [x] extended "no results" message doesn't appear anymore on submit
- [x] load aggregations in results query as well and use those, when available (and fix 2 spinners)
- [x] regression: filter isn't active on first filter click after a search

## Blocked by backend work
- [x] Video filter
- [x] surface format info in document results once backend supports it

## Crossbrowser sanity check
- [x] Chrome: looks good
- [x] Firefox: looks good
- [x] Safari: looks good
- [x] Chrome/Android: looks good
- [x] iOS/iPhoneX: looks good
- [x] IE11: minor issue (too many icons inside search input after submit, not launch blocking)
- [ ] Edge: TODO


uxengine:
@tpreusse Ready for a first pass. Please note:
- No need to comment on hardcoded translation strings, I'll follow up with a commit now that the dust starts settling.
- The preliminary code I mentioned is already removed.

components/Feed/index.js
```diff
@@ -62,6 +66,28 @@ const greetingSubscription = gql`
...
+    this.state = {
+      showSearch: false
+    }
+
+    this.onSearchClick = () => {
+      const showSearch = this.state.showSearch
+      this.setState({
+        showSearch: !showSearch
+      })
+      if (showSearch) {
```
tpreusse: you should always update the url. now it is inconsistent. If one clicks the icon anywhere but the feed they land on `feed?search=` if they click it while on the feed the url does not change. Correct would be `feed?search` (should work with the value `true`) always and `feed` when dismissing.

```diff
-if (showSearch) {
+Router.replaceRoute(
+  'feed',
+  showSearch? {search: true} : {},
+  { shallow: true }
+)
-}
```

components/Feed/index.js
```diff
@@ -70,6 +96,16 @@ class Feed extends Component {
     this.subscribe()
   }
 
+  componentWillReceiveProps (nextProps) {
+    const { query } = nextProps.url
+
+    if (query.hasOwnProperty('search') && !this.state.showSearch) {
```
tpreusse: afaik this shouldn't be needed once you handle it in the constructor since you already call set state in the onSearchClick handler.

components/Search/serialize.js
```diff
@@ -0,0 +1,45 @@
...
+  )
+}
+
+export const deserializeFilters = filtersString => {
+  return decodeURIComponent(filtersString)
+    .split('|')
+    .filter(filter => filter.split(':').length === 2)
+    .map(filter => ({ key: filter.split(':')[0], value: filter.split(':')[1] }))
+    .filter(filter => isSupportedFilter(filter))
+}
```
garamond: I think this would improve readability:
```
  return decodeURIComponent(filtersString)
    .split('|')
    .map(f => f.split(':'))
    .filter(f => f.length === 2)
    .map(([key, value]) => ({ key, value }))
    .filter(f => isSupportedFilter(f))
```

components/Search/Input.js
```diff
@@ -0,0 +1,105 @@
...
+          value={value}
+          ref={this.setFocusRef}
+          onChange={onChange}
+          onSearch={onSearch}
+          onReset={() => {}}
+          icon={
+            !value ? (
+              null
+            ) : allowSearch ? (
+              <Icon IconComponent={Search} onClick={onSearch} title={t('search/input/submit/aria')} />
```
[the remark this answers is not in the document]
garamond: Icon obscures logo on iPhone SE
iphone_se

components/Frame/Header.js
```diff
@@ -253,6 +289,25 @@ class Header extends Component {
...
+              onClick={(e) => {
+                e.preventDefault()
+                e.stopPropagation()
+                if (url.pathname === '/search') {
+                  window.scrollTo(0, 0)
+                } else {
+                  Router.pushRoute('search').then(() => window.scrollTo(0, 0))
+                }
+              }}>
+              <Search
```
garamond: Icon partly obscures logo on iPhone SE

iphone_se

uxengine:
Thanks, @garamond Quick wins fixed, I'll look into the two UI issues on Tuesday.

uxengine:
iPhone SE should also be OK now:
screen shot 2018-06-18 at 14 14 50

components/Frame/Header.js
```diff
@@ -253,6 +283,25 @@ class Header extends Component {
...
+              {...styles.search}
+              role='button'
+              title={t('header/nav/search/aria')}
+              onClick={(e) => {
+                e.preventDefault()
+                e.stopPropagation()
+                if (isSearchActive) {
+                  window.scrollTo(0, 0)
+                } else {
+                  Router.pushRoute('search').then(() => window.scrollTo(0, 0))
```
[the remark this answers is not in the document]
uxengine: Gave it a quick try and couldn't really get your suggestion working. Since I can't access the URL/path of the previous history entry before actually executing Router.back() or similar, how would I be able to implement the default origin? My guess is I'd need some app level state keeping a reference to lastPage or so.

components/Search/index.js
```diff
@@ -0,0 +1,284 @@
...
+        submittedQuery: this.state.searchQuery,
+        filterQuery: this.state.searchQuery,
+        filters: DEFAULT_FILTERS,
+        sort,
+        allowFocus: !this.state.isMobile
+      })
+      this.updateUrl(undefined, serializeSort(sort))
+      window._paq.push(['trackSiteSearch',
+        this.state.searchQuery,
+        false,
```
tpreusse: maybe add filters here as category? Otherwise seeing 0 results for something might not make sense without knowing the filters.

components/Frame/Header.js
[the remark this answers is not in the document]
garamond: You'd think that the Next router would provide access to the current history (at least through something like [`history.lenght`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/history.md)).  I agree, we probably need to track the history ourselves, or at least have something like a `hasNavigationOccurred` flag.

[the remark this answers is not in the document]
tpreusse: You can use a global variable in the header (or wherever you manage the icon).

See `beforeSearch` global being set in [`onSearch`](https://github.com/lobbywatch/website/blob/05ab20f96425b021c97ce6e7dba3044fb5524929/src/components/Frame/Header.js#L189) of lobbywatch.ch. The behaviour there is a bit more complex because it searches as you type and goes back by clearing the field.

components/Search/index.js
uxengine: Done. We're now also tracking the filters along with the count on filter click, while this one remains as is (filters are reset on search).

PR #104, opened 2018-05-09
fix: Ensure decoded email string

patrickvenetz:
This Merge Request ensures a decoded email string is printed in `Auth/SignIn` and `Auth/TokenAuthorization`.

It also exports `ensureDecodedEmail` in `lib/utils/base64u.js` as it is a shared, handy function. Before it was a function of `Auth/SignIn` component.

### DIFF
--- a/components/Auth/SignIn.js
+++ b/components/Auth/SignIn.js
@@ -5,7 +5,6 @@ import gql from 'graphql-tag'
 import {css} from 'glamor'
 import { Router } from '../../lib/routes'
 import withT from '../../lib/withT'
-import { ensureDecodedEmail } from '../../lib/utils/base64u'
 import isEmail from 'validator/lib/isEmail'
 import ErrorMessage from '../ErrorMessage'
 import RawHtmlElements from '../RawHtmlElements'
@@ -53,7 +52,7 @@ class SignIn extends Component {
   constructor (props) {
     super(props)
     this.state = {
-      email: ensureDecodedEmail(props.email) || '',
+      email: props.email || '',
       polling: false,
       loading: false,
       success: undefined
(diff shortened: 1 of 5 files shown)


patrickvenetz:
Got it, will do.

tpreusse:
@patrickvenetz I've thought about this some more and researched what you did in the backend.

My finding: Only the email in the verification url in [`backends/packages/auth/lib/signIn.js`](https://github.com/orbiting/backends/blob/298170b236d0df77692e44e06d764ce39d338641/packages/auth/lib/signIn.js#L79) is base64u encoded:
```js
  const verificationUrl =
    `${FRONTEND_BASE_URL}/mitteilung?` +
    querystring.stringify({
      type: 'token-authorization',
      email: encode(email),
      context,
      token
    })
```

Thats where we need it and where it makes sense. The corresponding place to decode in the front end is `pages/notifications.js`. _Side note: The routing of `mitteilung` to `notifications` is defined in `lib/routes.js`._

So let's not do a HoC, or any other fancy utils that spread around. Just a plain check directly in the page where the query is received:
```js
const safeEmail = email && (base64u.match(email)
  ? base64u.decode(email)
  : email
)
```

And then pass the `safeEmail` email down from there. Or did I miss any other actual case?

patrickvenetz:
> The corresponding place to decode in the front end is `pages/notifications.js`.

That was the "missing link".

Implementation now altered in dc8ef73, looks decent and it works.

A sideeffect is when user is redirected to another URL e.g. `invalid-token`, query parameter `email` will appear once again decoded — which is not a bad thing at all.

patrickvenetz:
> Or did I miss any other actual case?

Not that I know of (which is of little assurance). As far as I can tell [this](https://github.com/orbiting/backends/blob/master/packages/auth/lib/signIn.js#L75-L82) is the only generated URL with an email address in it.

Thanks for support @tpreusse.

PR #111, opened 2018-05-30
Post session message for native app

diegomura:
Notifies the native app for updates in the session state

### DIFF
--- a/lib/apollo/withMe.js
+++ b/lib/apollo/withMe.js
@@ -1,5 +1,6 @@
 import { graphql } from 'react-apollo'
 import gql from 'graphql-tag'
+import runInApp from '../utils/runInApp'
 
 export const meQuery = gql`
   query me {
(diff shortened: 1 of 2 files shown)


diegomura:
Thanks @tpreusse !
I was expecting fixes like these. I added the base post messaging communication, but I was aware of some of the issues you mention. 

*re serialization:* The message should be any type of data, but for some reason on Android devices I was not getting objects correctly 

*everything else*: agreed! Working on it

tpreusse:
Is there any urgent need to have this in prod? Otherwise should we create a app branch and merge all PR to there until things like UA ready to go master/prd?

diegomura:
No urgent at all,  but until we do both Android and iOS I'll release for you to test it won't work, since those are pointing to prod. 

Is there any staging environment we can use to test the apps? That would be ideal. I can easily make the apps point to there, and we can release these changes to also test them before prod

tpreusse:
Let's do a `app` branch with auto-deploy to a staging frontend pointing a prod. You could send a basic auth header in the web view right?

PR #112, opened 2018-05-30
Add iOS universal linking setup

diegomura:
Adds mandatory path to support universal linking on iOS apps. 
App ID to be determined

Refers to: https://github.com/orbiting/app/issues/4


diegomura:
Ready! The only caveat I see now is that when you go to `/apple-app-site-association`, the browser downloads the file. Before was getting rendered. This should work though

server/index.js
```diff
@@ -106,6 +107,11 @@ app.prepare().then(() => {
     return app.render(req, res, '/en', req.query)
   })
 
+  // iOS app universal links setup
+  server.use('/.well-known', express.static(
+    path.join(__dirname, '../static', '.well-known'), { extensions: ['json'] }
```
tpreusse: is the `extensions: ['json']` and `.json` file ending needed? Does iOS expect a content type json header?

diegomura: *I think* I tried to make the `apple-app-site-association` file without extension without succeeding on it, but I'll try again tomorrow because my memory is failing me on this time 😅 

Good catch about content-type. I didn't consider it when I did it. It has to be `Content-type: "application/pkcs7-mime"`. The current solution will return any file inside the `.well-known` dir, but maybe not all of them should have that content-type.

tpreusse: Does it not already work correctly? `extensions` from `express.static` is just a default extension to check.

If not do a specific route before the static dir (keep the static dir, we will need it later anyhow).
```
app.get('/.well-known/apple-app-site-association', (req, res) => {
  res.sendFile(fileName, {
    headers: {
      'Content-Type': 'application/json'
    }
  })
})
```

diegomura:
Ready! Now all I have to do is adding the appId

PR #113, opened 2018-05-31
App

tpreusse:
Monster PR introducing the following:
- apollo app link: forward request via postMessage to the app instead of http, thx @diegomura 
- app specific ui adaptions
- login via app: abstract and refactor poller and related text
- various improvements around sign in, thx @patrickvenetz 
- sub nav bar: developed for Feuilleton and already integrated in app, thx @uxengine 

ToDo:
- [x] review again
- [x] test
- [x] test again
- [x] and again
- [x] and again
- [ ] hide CC when in app
- [ ] test


pages/_document.js
```diff
@@ -36,7 +38,7 @@ export default class MyDocument extends Document {
         <Head>
           <meta
             name='viewport'
-            content='width=device-width,initial-scale=1'
+            content={`width=device-width,initial-scale=1 ${inNativeApp ? ',maximum-scale=1.0,user-scalable=0' : ''}`}
```
tpreusse: ends up being `width=device-width,initial-scale=1 ,maximum-scale=1.0,user-scalable=0` or `width=device-width,initial-scale=1 `. Small detail and probably works but IMO the space should go or be after every coma: `width=device-width, initial-scale=1, maximum-scale=1.0, user-scalable=0`.

diegomura: yes! you're right. Going to fix it

lib/apollo/withData.js
```diff
@@ -80,6 +81,12 @@ export default ComposedComponent => {
       }
     }
 
+    componentDidMount () {
```
diegomura: @patte @tpreusse @garamond 

I would like to discuss these lines.
The main purpose of this is having access to the `Router` instance in the app, so I can push routes client-side, without reloading the page as it is happening right now when you click on the native header ([here](https://github.com/orbiting/app/pull/100/files#diff-5a62d3a9fde9e73420c622a15da32f91R61))

However,  I _think_ this can be considered a security flaw, because this would be accessible from the console. It's not that someone can do anything I think (accessing a route that he does not have access). We can trigger this line only when app running in app also to make it more private.

But anyways, I would like to hear your thoughts. What I can say is that the app works MUCH nicer with this client-side redirection than without it.

patte: I leave that to the FE specialists to decide.

tpreusse:
Just for the record and to hopefully fix before merging (otherwise convert to issue):
We have inconsistent UX: discussion notifications can always enabled for app. Sign in preference not.
screen shot 2018-07-29 at 16 41 33

tpreusse:
In order to be able to release this we'll need to clarify what impact this has on our public user bases while the app is not available.

Currently I understand that: 
1. new app specific settings immediately become visible to all users
    - is this clear and desired? Initially my understanding was those settings would only appear after someone logged in with the app.
2. potential vulnerabilities related to `postMessage` without origin restriction
    - can `navigator.userAgend` be faked or are there cases where another site is open within the app that could steal data? Probably YouTube/Vimeo embeds.

I'll explore point 2 tonight.

lib/apollo/withData.js
```diff
@@ -63,7 +70,25 @@ export default ComposedComponent => {
...
+      // - browser only to ensure that it isn't shared between connections (which would be bad)
+      if (process.browser && this.props.headers) {
+        savedHeaders = this.props.headers
+      }
+    }
+
+    componentDidMount () {
+      // Make router accessible in a global way
+      // Useful in app to redirect page client-side
+      window.Router = Router
```
tpreusse: @diegomura still needed?

PR #115, opened 2018-05-31
Adjust frontend for native mobile app

diegomura:
### Description
Hides header and footer when frontend is rendered on mobile app.

### Pages changed
I just changed the pages that I could access from the app:

- Signin
- Profile
- Feed
- Article

I think the best is applying this change everywhere the `Frame` is rendered. Didn't do it yet because I would like to hear your thoughts

### Discussion
- I thought about using the context api to pass by the `renderedInApp` prop. the benefit here would be that we can just edit the `Frame` component and will work in all pages, making the code cleaner. Also, any other component can use that data to render, no matter how deep it is in the components tree. I didn't do it because I didn't see that you were using that API on the rest of the project. 

- Friendly suggestion: `Frame` JSX logic can be improved  by using some simple branch and render HOCs (such as the ones `recompose` provides).


components/Article/Page.js
```diff
@@ -368,6 +368,7 @@ class ArticlePage extends Component {
         formatColor={formatColor}
         audioSource={audioSource}
         audioCloseHandler={this.toggleAudio}
+        standalone={runningInApp}
```
diegomura: Couldn't find a better name for this prop. Seemed incorrect to me to make the `Frame` receive a prop called `runningInApp`, but I can change it easily if you want. `frameless` may be a valid option, or we can have two different props, one for `header` and other for `footer`

lib/nativeApp.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { HttpHeaderContext } from './apollo/withData'
+
+const matchUserAgent = value => value && value.match(/RepublikApp/)
+
+export const runInApp = process.browser && matchUserAgent(navigator.userAgent)
+  ? callback => callback()
+  : () => {}
+
+const withNativeSupport = WrappedComponent => props => {
```
tpreusse: Suggestion: `withInNativeApp`, rename the file to `withInNativeApp.js` if this is the default export

lib/nativeApp.js
```diff
@@ -0,0 +1,16 @@
...
+  return <HttpHeaderContext.Consumer>
+    {headers => <WrappedComponent runningInApp={matchUserAgent(headers.userAgent)} {...props} />}
```
tpreusse: `inNativeApp`

diegomura:
Awesome! Thanks for the changes 😄 

**re context**
Why not adding the `<HttpHeaderContext.Consumer>` straight in the `Frame` which is the component that needs the data? The reason I say this is because we would prevent having to pass down the `runningInApp` prop each time in all pages

**re branch**
I understand your point about `branch` and `recompose` in general, and makes sense. Despite the non-obvious args order, I like it because you end up with a much simpler and clearer JSX. There is no right or wrong in this point of course 😄

tpreusse:
*re context*
my thinking was you'd wrap `<Frame>` with `withNativeSupport` (aka `withInNativeApp`). I just did the minimal changes—getting the header context to work correctly. But yeah please do remove from pages and wrap directly.

diegomura:
@tpreusse Changes applied 👍

PR #118, opened 2018-06-07
Handle app authentication

diegomura:
This PR introduces:

- If user is in browser, signin flow is exactly as before
- If user is in app, the web sends a notification so it can handle the auth there
- Waits for the app to send the phrase, and starts polling as before

### DIFF
--- a/components/Auth/SignIn.js
+++ b/components/Auth/SignIn.js
@@ -61,11 +61,45 @@ const styles = {
 class SignIn extends Component {
   constructor (props) {
     super(props)
+
     this.state = {
       email: props.email || '',
       polling: false,
       loading: false,
       success: undefined
+    }
+
+    this.onFormSubmit = (event) => {
+      event.preventDefault()
+
+      const { loading, error, email } = this.state
+      const { signIn, context, acceptedConsents } = this.props
+
+      if (error) {
+        this.setState(() => ({ dirty: true }))
+        return
+      }
+
+      if (loading) {
+        return
+      }
+
+      this.setState(() => ({ loading: true }))
+
+      signIn(email, context, acceptedConsents)
+        .then(({data}) => {
+          this.setState(() => ({
+            polling: true,
+            loading: false,
+            phrase: data.signIn.phrase
+          }))
+        })
+        .catch(error => {
+          this.setState(() => ({
+            serverError: error,
+            loading: false
+          }))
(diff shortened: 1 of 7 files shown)


components/Auth/SignIn.js
```diff
@@ -61,12 +62,71 @@ const styles = {
...
+        return
+      }
+
+      if (loading) {
+        return
+      }
+
+      this.setState(() => ({ loading: true }))
+
+      runInApp(this.signInApp, this.signInBrowser)
```
diegomura: I think is cool to have a helper function that works as an 'if' statement egarding the user is in App or not. I don't know in this scenario if keeping the `runInApp` name is the best

diegomura:
Related with https://github.com/orbiting/app/pull/38

lib/withInNativeApp.js
```diff
@@ -3,9 +3,17 @@ import { HttpHeaderContext } from './apollo/withData'
 
 export const matchUserAgent = value => value && value.match(/RepublikApp/)
 
-export const runInApp = process.browser && matchUserAgent(navigator.userAgent)
-  ? callback => callback()
-  : () => {}
+export const runInApp = (appCallback, browserCallback) => {
```
tpreusse: There is a subtle difference between `withInNativeApp` and `runInApp`. `inNativeApp` is true while server rendering. `runInApp` does not run while server rendering for the app (important in the `withMe` case).

We should clean up the naming, make it more consitent and explicit. What do you think about adding `browser` or `webview` to `runInApp`?
```js
export const runningInNativeAppBrowser = process.browser && matchUserAgent(navigator.userAgent)
// or just inNativeAppBrowser? naming is soo hard
export const runInNativeAppBrowser = runningInNativeAppBrowser
  ? () => callback()
  : () => {}
```
_also notice the small difference of only evaluating `process.browser` and `navigator.userAgent` once, it does not change after the first evaluation_

(and then use the new `runningInNativeAppBrowser` export instead of a branch like function)

diegomura: I like it. Between `runningInNativeAppBrowser` and `inNativeAppBrowser` I like the most the latter.
However, IMO `AppBrowser` is a bit redundant. Maybe the most descriptive name would be `runninginNativeApp`. I can change it, but all options are good for me 😄

PR #120, opened 2018-06-08
Resolve graphql calls app side

diegomura:
@tpreusse Proof of concept of the idea you proposed about resolving graphql queries app side.

**What this does**
- Undo special signin handling
- Prevent using `HTTPLink` in app (I would then refactor the nested ternary conditionals) 
- Implement `AppLink` (name might change) that instead of resolving the request, it sends it to app
- Listens for response

It's the first time I implement an ApolloLink, so maybe you have enhancements  we can do in mind 😄 

I don't know yet which would be the best approach of correctly bind the message posting and response. Maybe you have thought about it.

Related to https://github.com/orbiting/app/pull/39

### DIFF
--- a/lib/apollo/appWorkerLink.js
+++ b/lib/apollo/appWorkerLink.js
@@ -1,6 +1,5 @@
 import { ApolloLink, Observable } from 'apollo-link'
 import { SubscriptionClient } from 'subscriptions-transport-ws'
-import { API_WS_URL } from '../constants'
 
 export const hasSubscriptionOperation = ({ query: { definitions } }) => (
   definitions.some(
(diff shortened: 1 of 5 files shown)


diegomura:
@tpreusse subscriptions ready, finally. Ready for review 😄

I extracted the new custom link functionality outside `initApollo`, but if you think there is a more structured way of doing it just tell me and I'll change it.

How it looks:
kapture 2018-06-12 at 15 41 31

lib/apollo/appWorkerLink.js
```diff
@@ -0,0 +1,120 @@
...
+    return this.subscriptionClient.request(operation)
+  }
+}
+
+// Create app worker link instance
+// Uses both promise or subscription links based on operation type
+export const createAppWorkerLink = () => {
+  const promiseWorkerLink = new PromiseWorkerLink()
+  const subscriptionWorkerLink = new SubscriptionWorkerLink({
+    uri: API_WS_URL,
```
tpreusse: drop all options: this does not connect to uri, doesn't do any reconnects or timeouts (afaik)

lib/apollo/appWorkerLink.js
```diff
@@ -0,0 +1,120 @@
...
+    this.protocol = protocol
+    this.readyState = WorkerInterface.OPEN
+  }
+
+  send (serializedMessage) {
+    window.postMessage(serializedMessage, '*')
+  }
+
+  set onmessage (fn) {
+    document.addEventListener('message', ({ data }) => {
```
tpreusse: how often is this called? should this also unregister a previous listener? (maybe track save the listener to the instance, remove it if present and create a new one with the new fn/callback)

lib/apollo/appWorkerLink.js
```diff
@@ -0,0 +1,120 @@
...
+      const d = JSON.parse(data)
+
+      if (d.type === 'start' || d.type === 'data') {
```
tpreusse: what about `stop`, `error` and `complete`? If error can happen with a specific query it should probably be handled

lib/apollo/appWorkerLink.js
```diff
@@ -0,0 +1,120 @@
...
+      })
+    })
+  }
+}
+
+// Apollo link implementation to resolve subscriptions via app worker
+class SubscriptionWorkerLink extends ApolloLink {
+  constructor ({ uri, options }) {
+    super()
+    this.subscriptionClient = new SubscriptionClient(uri, options, WorkerInterface)
```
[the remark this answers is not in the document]
diegomura: Agree. Good catch

PR #126, opened 2018-06-19
Paginated front

lukasbuenger:


components/Front/index.js
```diff
@@ -78,7 +116,33 @@ export default compose(
...
+            variables: {
+              first: 5,
+              after: data.front.children.pageInfo.endCursor
+            },
+            updateQuery: (previousResult = {}, { fetchMoreResult = {} }) => {
+              const previousSearch = previousResult.front.children || {}
+              const currentSearch = fetchMoreResult.front.children || {}
+              const previousNodes = previousSearch.nodes || []
+              const currentNodes = currentSearch.nodes || []
+              const res = {
```
patrickvenetz: Maybe `return` here already.

lukasbuenger: Yeah, thats remains of a log call. Will do. Thanks!

PR #127, opened 2018-06-19
Fix SSR

diegomura:
@tpreusse I found out that SSR was broken in `app` branch. This is _possibly_ the cause of the misbehaviors we observed in the past about server request. When I was requesting the feed page directly, the query was not being resolved server side, but client side. I went back into git history and I saw that this was introduced here: https://github.com/orbiting/republik-frontend/pull/113/commits/b354549afb595d0859ff67bbcc2c340ccfa0dddf 

I tried several fixes but all seemed to conclude that the new React context API was not behaving well when SSR. This PR implements the very same thing but using the last context API, which I confirmed it works.

I can show you this in practice tomorrow and discuss about it

### DIFF
--- a/lib/withInNativeApp.js
+++ b/lib/withInNativeApp.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { HttpHeaderContext } from './apollo/withData'
+import PropTypes from 'prop-types'
 
 export const matchUserAgent = value => value && value.match(/RepublikApp/)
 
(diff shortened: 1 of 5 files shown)


pages/feed.js
```diff
@@ -9,14 +9,14 @@ import withT from '../lib/withT'
 
 import { CDN_FRONTEND_BASE_URL } from '../lib/constants'
 
-const FeedPage = ({ url, me, t }) => {
+const FeedPage = ({ url, me, t, headers }) => {
```
tpreusse: not needed?

lib/withInNativeApp.js
```diff
@@ -1,13 +1,18 @@
...
 
 export const matchUserAgent = value => value && value.match(/RepublikApp/)
 
 export const inNativeAppBrowser = process.browser && matchUserAgent(navigator.userAgent)
 
 export const runInNativeAppBrowser = inNativeAppBrowser
   ? callback => callback()
-  : () => {}
+  : () => {
+    window.postMessage(JSON.stringify({
```
tpreusse: rm or put behind dev flag. otherwise we might end up forgetting

diegomura:
Thanks for your comments! Both things has to be gone. Sorry for that! They were just tests. I did this fix a bit in a hurry. Will fix it and merge it tomorrow

PR #128, opened 2018-06-19
Feed fixes

garamond:
* Improved edge detection
* Fix resize bug on mobile
* Add grey hairline to sticky header on mobile

jun-19-2018 22-52-59


components/Feed/StickySection.js
```diff
@@ -0,0 +1,143 @@
...
+
+  componentWillUnmount () {
+    window.removeEventListener('scroll', this.onScroll)
+    window.removeEventListener('resize', this.handleResize)
+  }
+
+  render () {
+    const { children, label } = this.props
+    const { sticky, isMobile } = this.state
+    const { width } = this.measure()
```
uxengine: Optional:
I was wondering if it's better not to call measure() on each render (since measure() might have already been called in onScroll and things add up), but instead in handleResize and cache the width in state.
Just a guess, don't have any data, but @tpreusse might have more background on whether or not my suggestion makes sense.

components/Feed/StickySection.js
```diff
@@ -0,0 +1,143 @@
...
+        {
+          children
+        }
+      </section>
+    )
+  }
+}
+
+StickySection.propTypes = {
+  spaceAfter: PropTypes.bool,
```
uxengine: I'd call this `hasSpaceAfter` to make it clear in the code above that it's a boolean and not some spacing value.

components/Feed/StickySection.js
```diff
@@ -0,0 +1,143 @@
...
+      width: SIDEBAR_WIDTH,
+      '& > div': {
+        width: SIDEBAR_WIDTH
+      }
+    }
+  }),
+  label: css({
+    padding: '8px 0',
+    borderTop: '1px solid #000',
+    position: 'relative',
```
uxengine: Not needed since you set the position explicitly in render()

components/Feed/StickySection.js
```diff
@@ -0,0 +1,143 @@
...
+    backgroundColor: '#fff',
+    width: '100%'
```
uxengine: This one's also explicitly set in render() and thus obsolete here.

components/Feed/StickySection.js
```diff
@@ -0,0 +1,143 @@
...
+    this.setSectionRef = (el) => { this.sectionRef = el }
+
+    this.onScroll = () => {
+      if (this.sectionRef) {
+        const { sticky, isMobile } = this.state
+        const { spaceAfter } = this.props
+        const y = window.pageYOffset + (isMobile ? HEADER_HEIGHT_MOBILE : HEADER_HEIGHT)
+        const { height, offset } = this.measure()
+        const nextSticky = (y > offset) && // scroll pos is above top of section
+          (offset + height + (spaceAfter ? STICKY_HEADER_HEIGHT : 0) > y) // scroll pos is below bottom
```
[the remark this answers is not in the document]
garamond: True

components/Feed/StickySection.js
garamond: Indeed

components/Feed/StickySection.js
garamond: You are absolutely right. I have limited calls to `measure` to `handleResize`.

components/Feed/StickySection.js
```diff
@@ -0,0 +1,127 @@
...
+          (offset + height + (hasSpaceAfter ? STICKY_HEADER_HEIGHT : 0) > y) // scroll pos is above bottom
+        if (sticky !== nextSticky) {
+          this.setState({sticky: nextSticky})
+        }
+      }
+    }
+
+    this.measure = () => {
+      const isMobile = window.innerWidth < mediaQueries.lBreakPoint
+      const isSmall = window.innerWidth < mediaQueries.mBreakPoint
```
uxengine: Cool, I just wanted to comment on the jumping header in non-floating mode on desktop, but I've seen you just fixed it with introducing `isSmall`.

I just think the naming of `isMobile` could be misleading, since the header and the rest of the page are still in desktop mode.

PR #132, opened 2018-06-26
feat(ActionBar): move all share actions behind a share icon+overlay

uxengine:
- We're starting to have space issues on smaller screens – we basically have too many share icons and need to be able to scale
- Aligned with how upcoming mobile app will treat share icons (one icon + native share overlay)
- Also rename Share.js to ActionBar/index.js to reflect how it's actually used.

Before (iPhoneSE/5):
screen shot 2018-06-26 at 09 49 25

After (iPhoneSE/5):
screen shot 2018-06-26 at 10 43 44

Overlay (iPhoneSE/5):
screen shot 2018-06-26 at 09 51 02

Overlay (iPhone8):
screen shot 2018-06-26 at 09 51 51

Overlay (iPad):
screen shot 2018-06-26 at 09 52 07

Overlay (Desktop):
screen shot 2018-06-26 at 09 52 38

Profile share overlay (Desktop):
screen shot 2018-06-26 at 09 53 03


uxengine:
@clarajeanne Please checkout the visual changes and LMK if you have any concerns.

garamond:
The header of the sharing menu seems to have a different height than the main header.

localhost_3010_2018_06_04_bank-asya-dan-tmsf-ye-360-milyon-lira-oedeme iphone x

garamond:
I think the close buttons should be identical.

localhost_3010_2018_06_04_bank-asya-dan-tmsf-ye-360-milyon-lira-oedeme iphone 5_se

uxengine:
@garamond Good catch, both is also true for the already existing PDF overlay which uses the same styleguide component. Don't think it's a blocking issue, but I agree the styleguide should be updated to match our defaults.

garamond:
IMO the menu should close once the article has been shared (i.e. one of the `IconLink`s has been clicked)?

components/ActionBar/index.js
```diff
@@ -0,0 +1,146 @@
...
+            title={shareOverlayTitle || t('article/actionbar/share')}
+            tweet={tweet}
+            emailSubject={emailSubject}
+            emailBody={emailBody}
+            emailAttachUrl={emailAttachUrl} />
+        )
+        }
+        <span {...styles.buttonGroup}>
+          {icons
+            .filter(Boolean)
```
garamond: Neat shortcut!

components/IconLink.js
```diff
@@ -82,14 +107,15 @@ const IconLink = ({
...
+  onClick,
+  stacked
 }) => {
   const Icon = ICONS[icon]
 
   return (
     <a
       {...styles.link}
-      {...(mobileOnly ? styles.mobileOnly : {})}
+      {...(getExtraStyles(mobileOnly, stacked))}
```
garamond: It's just a small thing, but I think the conditional styles should me merged into the `styles` object (with `css` already applied), and the test should be performed inline.

    {...merge(
       stacked && styles.stacked, 
       mobileOnly && styles.mobileOnly
    )}

To my eyes this would improve readability.

uxengine: @garamond I agree with the readability aspect, but we tend to avoid merge() at render time for performance reasons.

components/ActionBar/index.js
```diff
@@ -0,0 +1,146 @@
...
+      shareOverlayTitle
+    } = this.props
+    const { showShareOverlay } = this.state
+
+    const icons = [
+      {
+        icon: 'share',
+        href: url,
+        onClick: e => {
+          if (e.currentTarget.nodeName === 'A' &&
```
tpreusse: Rm this code. Trying to open the share overlay in a new window is not supported. It should `e.preventDefault()` and show the overlay in the current window.

PR #136, opened 2018-07-03
update crew section on imprint to gsheet-based setup plus community-like design

uxengine:
Depends on https://github.com/orbiting/backends/pull/66

- This will allow to manage the crew information via a Google Sheet without IT involvement.
- Instead of the text list, we re-use the profile image tiles from the community page.
- Also tweaking the breakpoints a bit to get reasonable tile sizes with the longer name+title strings.

Loading is done in two steps:
1. Load gsheet data containing the group/subgroup setup and userIds + titles per  group context
2. Load user data (i.e. profile image) for all userIds received in step.

## Previews
(please disregard the missing images on my local instance)

Desktop:
screen shot 2018-07-03 at 17 34 31

Mobile:
screen shot 2018-07-03 at 17 34 47


components/Imprint/Employees.js
```diff
@@ -0,0 +1,199 @@
...
+  })
+)(({ employeeGroups, data: { loading, error, search } }) => {
+  return (
+    <Loader
+      loading={loading}
+      error={error}
+      render={() => {
+        const { nodes } = search
+
+        const userById = [...nodes].reduce((map, node) => {
```
garamond: I think `nest` would come in handy here as well…

components/Imprint/Employees.js
```diff
@@ -0,0 +1,199 @@
...
+              <section {...styles.group} key={i}>
+                <H2 {...styles.groupHeading}>{groupKey}</H2>
+                {!!employeeGroups[groupKey] &&
+                  !employeeGroups[groupKey].group &&
+                  Object.keys(employeeGroups[groupKey]) &&
+                  Object.keys(employeeGroups[groupKey]).map((subgroupKey, i) => (
+                    <section {...styles.subgroup} key={i}>
+                      <H3 {...styles.subgroupHeading}>{subgroupKey}</H3>
+                      {!!employeeGroups[groupKey][subgroupKey].group &&
+                        employeeGroups[groupKey][subgroupKey].group.map((employee, i) => (
```
garamond: [`nest.entries()`](https://github.com/d3/d3-collection#nest_entries) gives you a mappable array of nested key/value pairs

components/Imprint/Employees.js
```diff
@@ -0,0 +1,199 @@
...
+          })
+        )
+      ].filter(Boolean)
+
+      // Create a nested object with keys being group names and values being either:
+      // - a {list: [...]} object containing an array of employees, or
+      // - keys of subgroups names pointing to a {list: [...]} object
+      const firstLevelGroups = groupBy(employees, 'group')
+      let employeeGroups = {}
+      for (var property in firstLevelGroups) {
```
garamond: I think the same could achieved with: 

```
nest()
  .key(d => d['group'])
  .key(d => d['subgroup'] || 'group')
  .object(employees)
```

components/Imprint/Employees.js
```diff
@@ -0,0 +1,199 @@
...
+            portrait
+            username
+          }
+        }
+      }
+    }
+  }
+`
+
+const groupBy = (array, key) => {
```
garamond: I propose to use [`nest`](https://github.com/d3/d3-collection#nests) from `d3-collection` for grouping lists – it's more succinct and less error-prone.

uxengine:
@garamond Thanks for the review. All d3'ified now.

PR #138, opened 2018-07-06
add link to upgrade to yearly membership

lukasbuenger:
image

Wording subject to change.


patte:
The UX of this is insufficient, in my opinion. I think it's not comprehensible to users what's going to happen, they are just thrown to the buy page, without any instructions or explanations. I suggest, that when clicking on that link, a component folds out that explains: "When clicking on the following link you will be redirected to the buying page of a yearly membership. We'll automatically cancel your monthly membership the moment you bought a yearly...". The info text is followed by a button "Buy Yearly". What do you think about it @lukasbuenger @uxengine?

uxengine:
@patte I support the idea of giving more context. Maybe we can find a simpler solution, e.g. just one sentence explaining things and linking off, e.g.
`Wenn Sie auf ein <a>Jahresabo upgraden</a>, beenden wir automatisch Ihr Monatsabo.`

PR #139, opened 2018-07-06
WIP: authenticate by app

patte:
these images just show the usage in the app, the code for this feature is the same as for EMAIL_TOKEN.
screen shot 2018-07-07 at 01 02 02
screen shot 2018-07-07 at 01 02 26

new feature: denySession
screenshot from 2018-07-27 18-32-57

don't show 'close Window' messages inNativeApp or if not needed also in browser
screenshot from 2018-07-27 18-49-47
screen shot 2018-07-07 at 03 24 57

choose the preferred first factor:
screenshot from 2018-07-26 15-03-09
screenshot from 2018-07-26 15-03-20


patte:
@uxengine can you help me with the styling? I think the "Zugriff ablehnen" button could be orange? What's our way of handling this?

uxengine:
@patte I would expect the button to be green (Republik primary color). You need to define SG_COLORS in your local .env for that, styleguide's primary color is blue. Will send you via chat.

philippvonessen:
as discussed with patte, i would vote for less information on the mobile screen. some snapshot from  google...

Screen 1 on desktop
bildschirmfoto 2018-07-09 um 16 08 09

Screen 2 «Alternatives» on desktop
bildschirmfoto 2018-07-09 um 16 07 39

Screen 3 on mobile
screenshot_2018-07-09-16-08-36

components/Auth/TokenAuthorization.js
```diff
@@ -14,19 +15,30 @@ import ErrorMessage from '../ErrorMessage'
...
 const { P } = Interaction
 
 const goTo = (type, email) => Router.replaceRoute(
   'notifications',
   { type, email, context: 'authorization' }
 )
 
-const shouldAutoAuthorize = ({ error, target }) => {
+const shouldAutoAuthorize = ({ error, target, noAutoAuthorize }) => {
+  console.log({noAutoAuthorize})
```
uxengine: Remove

pages/notifications.js
```diff
@@ -80,6 +86,8 @@ const Page = withT(({ url: { query, query: { context, token } }, t }) => {
     content = <TokenAuthorization
       email={email}
       token={token}
+      tokenType={tokenType || 'EMAIL_TOKEN'}
```
uxengine: Optional:
I've introduced `DEFAULT_TOKEN_TYPE` in Account/AuthSettings.js and use it for a similar fallback there - so you might wanna re-use that constant (or move it to e.g. components/constants.js).

pages/notifications.js
```diff
@@ -148,10 +167,19 @@ const Page = withT(({ url: { query, query: { context, token } }, t }) => {
               )), () => ' – ')}
             </P>
           )}
+          {!hasCurtain && buttonLinks.length > 0 && buttonLinks.map((link, i) => (
```
uxengine: Looks like buttonLinks is guaranteed to be an array, so you wouldn't need the extra check `buttonLinks.length > 0` because an empty array simply woudn't map anything.

components/Auth/TokenAuthorization.js
```diff
@@ -78,7 +114,8 @@ class TokenAuthorization extends Component {
       echo,
       email,
       error,
-      loading
+      loading,
+      noAutoAuthorize
```
tpreusse: What is this?

components/Account/AuthSettings.js
```diff
@@ -0,0 +1,149 @@
...
+          }
+        })
+    })
+  }),
+  graphql(query, {
+    props: ({ data, errors }) => ({
+      data,
+      loading: data.loading || !data.me,
+      error: data.error,
+      me: data.loading ? undefined : data.me
```
tpreusse: it is misleading to call this `me`. Call it `authSettings` or directly map the two variables (`enabledFirstFactors` and `preferredFirstFactor`)

components/Auth/TokenAuthorization.js
```diff
@@ -14,19 +15,29 @@ import ErrorMessage from '../ErrorMessage'
...
 
-const shouldAutoAuthorize = ({ error, target }) => {
+const shouldAutoAuthorize = ({ error, target, noAutoAuthorize }) => {
   return (
     !error &&
     target &&
     target.session.isCurrent &&
-    !target.requiredConsents.length
+    !target.requiredConsents.length &&
+    noAutoAuthorize === undefined
```
tpreusse: why not `!noAutoAuthorize`?

components/Auth/TokenAuthorization.js
patte: this is a new query param, going to be sent in the updated signIn-Email's notice link: "if you didn't request this, deny it here". It ensures, that no autoAuthorize is started.

components/Account/AuthSettings.js
patte: @uxengine: fyi 
@tpreusse: as he is in holidays, I would appreciate if you can clean it up.

patte:
> Why did you add sign in via app to the preview mail? What is the use case? Looks like unnecessary complexity.
It's an edge case, but users with an app enrolled can also request a preview. The backend would correctly use the `App` signIn strategy, but the preview auth would show the email notice. That's why I made Auth/SignIn and preview/SignIn behave the same.

tpreusse:
@patte We have similar edge cases for FAQ and claim. If we want to handle those we need to abstract polling and support that everywhere. Another solution would be to enforce sign in via email for `claim`, `preview` and a new `faq` context in the backend. Possibly `pledge` too.

tpreusse:
In order to be able to release this we'll need to clarify all edge cases: `faq`, `claim` and `pledge`. I'm not able to think through all those cases today, my current understanding is that `faq` and `claim` is similarly broken as `preview` was. `pledge` might be fine but most likely the wording is wrong if someone authorises in via app.

PR #147, opened 2018-07-24
Article gallery

garamond:
Add a gallery component to show all photos of an article.

 * Click on any image to enlarge it
 * Click/tap on gallery center to hide/show caption and header
 * Swipe/click left/right to navigate
 * Swipe down to exit

gallery

### DIFF
--- a/components/Article/Page.js
+++ b/components/Article/Page.js
@@ -18,6 +18,8 @@ import Feed from '../Feed/Format'
 import StatusError from '../StatusError'
 import SSRCachingBoundary, { webpCacheKey } from '../SSRCachingBoundary'
 import withMembership from '../Auth/withMembership'
+import { Gallery } from '@project-r/styleguide/lib/components/Gallery'
+import PropTypes from 'prop-types'
 
 import {
   colors,
@@ -35,6 +37,7 @@ import createFormatSchema from '@project-r/styleguide/lib/templates/Format'
 import createDossierSchema from '@project-r/styleguide/lib/templates/Dossier'
 import createDiscussionSchema from '@project-r/styleguide/lib/templates/Discussion'
 import createNewsletterSchema from '@project-r/styleguide/lib/templates/EditorialNewsletter/web'
+import get from 'lodash.get'
 
 const schemaCreators = {
   editorial: createArticleSchema,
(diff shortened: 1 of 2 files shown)


components/Article/Page.js
```diff
@@ -180,6 +187,9 @@ class ArticlePage extends Component {
       secondaryNavExpanded: false,
       showSecondary: false,
       showAudioPlayer: false,
+      gallery: {
+        show: false
```
uxengine: Maybe it's just me, but I like to use the initial state to document potential state changes, so I would already include the startItem here:

```
gallery: {
  show: false,
  startItem: null
}
```

components/Article/Page.js
```diff
@@ -240,6 +250,74 @@ class ArticlePage extends Component {
         secondaryNavExpanded: expanded
       })
     }
+
+    this.getGalleryItems = memoize(() => {
+      const { data: { article } } = this.props
+      const shouldInclude = (el) =>
+        get(el, 'identifier', '') === 'FIGURE' && get(el, 'data.excludeFromGallery', false) !== true
+      function findFigures (node, acc = []) {
```
uxengine: Wearing the ES6 hat:
`const findFigures = (node, acc = []) => {`

components/Article/Page.js
```diff
@@ -240,6 +250,74 @@ class ArticlePage extends Component {
...
+          caption,
+          credit
+        }
+      }
+      return findFigures(article.content)
+        .map(getImageProps)
+        .filter(i => imageSizeInfo(i.src).width > 600)
+    })
+
+    this.renderGallery = () => {
```
uxengine: I wonder whether all new code here (including context) should be moved to a separate file to keep file length down. But I might not see any complications out of this yet.

garamond: You are right, of course. It's much nicer to encapsulate the whole thing in a separate component!

components/Article/ArticleGallery.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react'
```
uxengine: Just for consistency: We typically do
`import React, { Fragment } from 'react'`
so we don't have to namespace the component further down.

components/Article/ArticleGallery.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    this.getGalleryItems = memoize(() => {
+      const { article } = this.props
+      return findFigures(article.content)
+        .map(getImageProps)
+        .filter(i => imageSizeInfo(i.src).width > 600)
+    })
+
+    this.renderGallery = () => {
```
uxengine: This render code could live directly inside `render()`, especially now that it's a separate and small component.

PR #148, opened 2018-07-26
Revert "Article image gallery"

garamond:


garamond:
Yes, wanted to remove that anyway

PR #151, opened 2018-07-30
fix: Remove instant error feedback when entering email address

patrickvenetz:
This Pull Request alters all known form fields in which you enter an email address as follows: It won't validate until you hit enter or push a button but no more on blurring.

This [change request](https://trello.com/c/Pa0QWHMS/69-34-login-optimierung) originates from cases Careteam allegedly ran into.

To keep it somewhat consistent, not only requested login form field was updated but all fields in which you add an email address.


patrickvenetz:
Careteam is researching cases over the course of the next couple of days.

PR #152, opened 2018-07-31
Display error message if cookies appear disabled

patrickvenetz:
This Pull Request show an error message and some explanatory text on sign-in forms and offers page if cookies appear to disabled.

It determines that checking `navigator.cookieEnabled`.

As of July 31, 2018, wording is not yet settled.

Part of [Login-Optimierung](https://trello.com/c/Pa0QWHMS/69-34-login-optimierung).

### SignIn component inside hamburger menu
signin-hamburger-menu

### SignIn component on `/signin` page
signin-page

### Not rendered submit form on FAQ pa
Not rendered submit form on FAQ page

### Hidden offers page
offers


patrickvenetz:
Adopted.

components/Auth/SignIn.js
```diff
@@ -65,17 +65,25 @@ class SignIn extends Component {
...
       polling: false,
       loading: false,
-      success: undefined
+      success: undefined,
+      cookiesEnabled: true
+    }
+  }
+
+  componentDidMount () {
+    if (navigator) {
```
tpreusse: This check is a bit pointless, running it in node would explode:
```
> if (navigator) console.log(123)
ReferenceError: navigator is not defined
```

But `componentDidMount` is always only run in the browser. No need for checking weather a navigator object exists here.

components/Pledge/Form.js
```diff
@@ -129,19 +133,36 @@ class Pledge extends Component {
   }
   componentDidMount () {
     this.checkUserFields(this.props)
+
+    if (navigator) {
+      this.setState({ cookiesEnabled: navigator.cookieEnabled })
```
tpreusse: The whole payment flow works without cookies from our side (e.g. a stripe purchase is possible without cookies, not sure about the others). I'd prefer to keep allowing this. Please remove from pledge form again. I'll make sure that this cookie error message becomes visible in all flows eventually (through a upcoming poller abstraction).

patrickvenetz: Had same thought train going here.

Here's why I believe we should not allow a user walk through the payment process: While the payment flow works, the sing-in and confirmation procedure hereafter does not. As I understand, you can not complete the payment process fully nor use Republik hereafter if you haven't cookies enabled.

To prevent wrong notions, it feels more user-friendly (...) to not show checkout page.

Meanwhile, there would be that lovely option of listing what we would have to offer if you were inclined to enabled cookies: List pledges. This might still be option. I refrained from that due to my rudimentary frontend knowledge, speed of implementing it and "income/expense ratio".

components/Auth/SignIn.js
patrickvenetz: Will do.

patrickvenetz:
(I appreciate the time your investing to help me here.)

PR #154, opened 2018-08-02
New front render page and route

lukasbuenger:
Adds a new rendering page for non-index front documents and a route that points at it.

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -45,5 +45,6 @@ routes
   .add('format', '/format/:slug', 'article')
   .add('dossier', '/dossier/:slug', 'article')
   .add('article', '/:year(\\d{4})/:month(\\d{2})/:day(\\d{2})/:slug/:suffix(diskussion)?')
+  .add('front', '/:slug', 'front')
 
 module.exports = routes
--- a/pages/front.js
+++ b/pages/front.js
@@ -1,0 +1,16 @@
+import React from 'react'
+import { compose } from 'react-apollo'
+import Front from '../components/Front'
+import withData from '../lib/apollo/withData'
+import withT from '../lib/withT'
+import { enforceMembership } from '../components/Auth/withMembership'
+
+const FrontPage = ({ url, t, isMember, headers }) => {
+  return <Front url={url} headers={headers} />
+}
+
+export default compose(
+  withData,
+  enforceMembership,
+  withT
+)(FrontPage)


pages/front.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { compose } from 'react-apollo'
+import Front from '../components/Front'
+import withData from '../lib/apollo/withData'
+import { enforceMembership } from '../components/Auth/withMembership'
+
+const FrontPage = ({ url, headers }) => {
+  return <Front url={url} headers={headers} />
```
tpreusse: Path is hard-coded in [`<Front>`](https://github.com/orbiting/republik-frontend/blob/d78b56a67612e695e03d6375e475f91e3ee7387e/components/Front/index.js#L108). You'll need to make that dynamic, use [`url.asPath.split('?')[0]`](https://github.com/orbiting/republik-frontend/blob/d78b56a67612e695e03d6375e475f91e3ee7387e/components/Article/Page.js#L430) and please test carefully.

lib/routes.js
```diff
@@ -45,5 +45,6 @@ routes
   .add('format', '/format/:slug', 'article')
   .add('dossier', '/dossier/:slug', 'article')
   .add('article', '/:year(\\d{4})/:month(\\d{2})/:day(\\d{2})/:slug/:suffix(diskussion)?')
+  .add('front', '/:slug', 'front')
```
tpreusse: `.add('front', '/:slug')` should suffice and is more consistent with the rest.

components/Front/index.js
```diff
@@ -103,9 +103,9 @@ class Front extends Component {
 export default compose(
   withT,
   graphql(getDocument, {
-    options: () => ({
+    options: props => ({
       variables: {
-        path: '/',
+        path: props.path || '/',
```
tpreusse: why not use `props.url.asPath.split('?')[0]` directly here? Would make it unnecessary to introduce a new path props and work just like the article page.

pages/front.js
```diff
@@ -0,0 +1,14 @@
...
+import { enforceMembership } from '../components/Auth/withMembership'
+
+const FrontPage = ({ url, headers }) => {
+  return <Front url={url} headers={headers} path={url.asPath.split('?')[0]} />
+}
+
+export default compose(
+  withData,
+  enforceMembership
+)(FrontPage)
```
tpreusse: You could directly use the `Front` component here if you ready the asPath from the url object inside the component.

PR #159, opened 2018-08-09
WIP feuilleton

uxengine:


tpreusse:
@uxengine Lukas recently created a catch all front rule: `add('front', '/:slug')`. If you create a front document with the path feuilleton it should just work without any code.

PR #166, opened 2018-08-20
feat(Account): Access compontents

patrickvenetz:
tba


components/Account/Access/Campaigns/Form.js
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    this.onChangeEmail = (event, value, shouldValidate) => {
+      event.preventDefault()
+
+      this.setState(() => ({
+        value,
+        error: (
+          value.trim().length <= 0 &&
+          'Keine E-Mail-Adresse angegeben'
```
tpreusse: Make sure to use `t`. Also consider that the error replaced the label, I prefer following style when possible:  
`E-Mail-Adresse`
`E-Mail-Adresse fehlt`  
`E-Mail-Adresse ungültig`  
_keeping the label in the same place at the beginning_

components/Account/Access/Campaigns/Grant.js
```diff
@@ -0,0 +1,25 @@
...
+
+const dayFormat = timeFormat('%e. %B %Y')
+
+class Grants extends Component {
+  render () {
+    const { grant, revokeAccess } = this.props
+
+    return (
+      <Item key={grant.id}>
+        an <strong>{grant.email}</strong><br />
```
tpreusse: Avoid native `strong`. This leads to [faux bold](https://alistapart.com/article/say-no-to-faux-bold). Use [`Interaction.Emphasis`](https://styleguide.republik.ch/typography#emphasis-1)

components/Account/Access/Campaigns/Grant.js
```diff
@@ -0,0 +1,25 @@
...
+
+import { Item } from '../../../List'
+
+import { timeFormat } from '../../../../lib/utils/format'
+
+import Revoke from './Revoke'
+
+const dayFormat = timeFormat('%e. %B %Y')
+
+class Grants extends Component {
```
tpreusse: if you don't plan to add state here us a function instead of a class:
```js
const Grants = ({ grant, revokeAccess }) => (
  <MyJSX />
)
```

components/Account/Access/Campaigns/Grants.js
```diff
@@ -0,0 +1,35 @@
...
+    const { campaign, revokeAccess } = this.props
+
+    if (campaign.grants.length === 0) {
+      return null
+    }
+
+    return (
+      <Fragment>
+        {campaign.slots.used > 1
+          ? <H3 style={{marginTop: 30}}>Vergebene Zugriffe</H3>
```
tpreusse: move ternary inside a t statement – `t(`my/key/space/${true ? 'A' : 'B'}`)`

components/Account/Access/Campaigns/Form.js
```diff
@@ -0,0 +1,168 @@
...
+          !isEmail(value) &&
+          'Keine gültige E-Mail-Adresse erkannt'
+        ),
+        dirty: {
+          email: shouldValidate
+        }
+      }))
+    }
+
+    this.onSubmit = async (event) => {
```
tpreusse: unneeded async (uses promise syntax inside, most FE code currently uses Promises no need to change it)

components/Account/Access/Campaigns/Form.js
```diff
@@ -0,0 +1,168 @@
...
+          <Label>Alle verfügbaren Zugriffe vergeben.</Label>
+        )
+      }
+
+      return null
+    }
+
+    const {
+      value,
+      error = false,
```
tpreusse: This is redundant and unnecessary. Redundant because already false in initial state via construction. Unnecessary because error undefined would work fine.

components/Account/Access/Campaigns/Form.js
```diff
@@ -0,0 +1,160 @@
...
+    super(props)
+
+    this.state = {
+      email: '',
+      error: false,
+      isMutating: false,
+      hideForm: true,
+      mutationError: null,
+      isUsed: props.campaign.slots.used > 0,
+      dirty: {}
```
tpreusse: Since you only have one field: no need for an object. Start with `dirty: false` or `dirty: undefined` or just leave it of of the initial state (implicit undefined).

components/Account/Access/Campaigns/Form.js
```diff
@@ -0,0 +1,160 @@
...
+                {t.pluralize(
+                  'Account/Access/Campaigns/Form/title',
+                  { count: campaign.slots.used }
+                )}
+              </H3>
+              <Field
+                name='email'
+                type='email'
+                label={t('Account/Access/Campaigns/Form/input/email/label')}
+                error={error}
```
tpreusse: `error={dirty && error}`

if you keep the dirty object: `error={dirty.email && error}`

components/Account/Access/Campaigns/Form.js
```diff
@@ -0,0 +1,160 @@
...
+        ),
+        dirty: {
+          email: shouldValidate
+        }
+      }))
+    }
+
+    this.onSubmit = (event) => {
+      event.preventDefault()
+
```
tpreusse: prevent onSubmit when there is a client error:
```js
if (this.state.error) {
  this.setState({dirty: true}) // or dirty: {email: true}
  return
}
```

components/Account/Access/Grants.js
```diff
@@ -0,0 +1,59 @@
...
+        <P>{t('Account/Access/Grants/explanation')}</P>
+        {accessGrants.map((grant, i) => (
+          <P key={`grant-${grant.id}`}>
+            {t('Account/Access/Grants/grant', {
+              grantee: grant.grantee.name || grant.grantee.email,
+              endAt: dayFormat(new Date(grant.endAt))
+            })}
+          </P>
+        ))}
+        {!inNativeIOSApp &&
```
tpreusse: 👏

components/Account/NewsletterSubscriptions.js
```diff
@@ -70,11 +70,8 @@ class NewsletterSubscriptions extends Component {
               {subscriptions.map(({ name, subscribed, isEligible }) => (
                 <p key={name}>
                   <Checkbox
-                    checked={subscribed}
-                    disabled={
-                      (!isEligible && !subscribed) ||
-                      mutating[name]
-                    }
+                    checked={subscribed && isEligible}
+                    disabled={!isEligible || mutating[name]}
```
tpreusse: Why? Once the membership resumes they would automatically be re-subscripted. Current logic is that they can unsub while not eligible but can't resub. After this change people could come under the impression that we lost their preference.

components/Account/index.js
```diff
@@ -30,87 +35,136 @@ import PaymentSources from './PaymentSources'
...
+    <div style={{marginBottom: 80}}>
+      <a {...styles.accountAnchor} id={id} />
+      {children}
+    </div>
+  )
+}
+
+class Account extends Component {
+  componentDidMount () {
+    const match = /#(\w+)$/m.exec(this.props.router.asPath)
```
tpreusse: `location.hash` should do:
`window.location.hash.substr(1)` -> `''` or `'teilen'`

components/Frame/Popover/Nav.js
```diff
@@ -197,6 +197,28 @@ const Nav = ({ me, url, closeHandler, children, t, inNativeApp, inNativeIOSApp,
               <br />
             </Fragment>
           )}
+          {!!me && (
```
tpreusse: Currently quite unbalanced:
screen shot 2018-09-02 at 14 33 42

Consider moving to the left side underneath `Konto`. And use better label: `Teilen` is too generic. Maybe `Zugriff teilen`.

components/Account/Access/Campaigns/index.js
```diff
@@ -0,0 +1,75 @@
+import { compose, graphql } from 'react-apollo'
+import gql from 'graphql-tag'
+
+import Campaign from './Campaign'
+
+import query from '../../belongingsQuery'
+
+const Campaigns = ({ accessCampaigns, grantAccess, revokeAccess, state }) => {
+  return accessCampaigns.length > 0 && accessCampaigns.map((campaign, key) => {
```
tpreusse: why is `accessCampaigns.length > 0` necessary? Returning `[]` or `false` from a component results in the same dom.

components/Account/Access/Campaigns/index.js
```diff
@@ -0,0 +1,75 @@
+import { compose, graphql } from 'react-apollo'
+import gql from 'graphql-tag'
+
+import Campaign from './Campaign'
+
+import query from '../../belongingsQuery'
+
+const Campaigns = ({ accessCampaigns, grantAccess, revokeAccess, state }) => {
```
tpreusse: is `state` prop used? A bit unfortunate naming as it is then spread as props onto `<Campaign />`

components/Frame/Popover/Nav.js
tpreusse: better label idea: `Republik teilen`

components/Account/NewsletterSubscriptions.js
patrickvenetz: I found UI better to describe current state of your subscription, hereby moving its meaning from "subscribed, but not in segment = no mail" to "no subscription = no mail" and by losing eligibility losing subscription, and not accounting for loss (and restoring) preferences here.

So much for why part.

Reverting however to previous logic.

tpreusse: I agree that we can do better here. We can tackle this separately later.

components/Account/Access/Campaigns/index.js
patrickvenetz: It isn't.

components/Account/Access/Campaigns/index.js
patrickvenetz: It is not used.

components/Account/Access/Campaigns/Form.js
```diff
@@ -0,0 +1,166 @@
...
+      this.setState(() => ({
+        value,
+        error: (
+          value.trim().length <= 0 &&
+          this.props.t('Account/Access/Campaigns/Form/input/email/missing')
+        ) || (
+          !isEmail(value) &&
+          this.props.t('Account/Access/Campaigns/Form/input/email/invalid')
+        ),
+        dirty: {
```
tpreusse: should be `dirty: shouldValidate`

patrickvenetz:
Changes adopted. Please revisit.

components/Account/belongingsQuery.js
```diff
@@ -32,6 +32,32 @@ query myBelongings {
         name
       }
     }
+    accessGrants {
+      endAt
+      campaign {
+        title
+        description
+      }
+      grantee {
```
tpreusse: you potentially want to change this to `granteeName` after backend revisions

components/Account/Access/Grants.js
```diff
@@ -0,0 +1,59 @@
...
+
+const AccessGrants = ({ accessGrants, inNativeIOSApp, t }) => {
+  return accessGrants.length > 0 && (
+    <Box>
+      <MainContainer>
+        <P>{t('Account/Access/Grants/explanation')}</P>
+        {accessGrants.map((grant, i) => (
+          <P key={`grant-${grant.id}`}>
+            {t('Account/Access/Grants/grant', {
+              grantee: grant.grantee.name || grant.grantee.email,
```
tpreusse: After backend revisions: `grant.granteeName`

PR #171, opened 2018-08-27
Add label with frequency in Account/NewsletterSubscription

patrickvenetz:
This Pull Request renders a label inside Newsletter checkbox. Label contains "frequency" of a newsletter, how often or when it is published.

frequency label


patrickvenetz:
Albeit not part of this Pull Request, upcoming changes towards "Feuilleton", frequency daily newsletter changes and "Feuilleton" newsletter steps into it's place and should eventually look like this:

frequency, including feuilleton newsletter

patrickvenetz:
Oh no, it does:

iPhone 5

How could I prevent that?

PR #175, opened 2018-08-28
App Round Two

tpreusse:
Changes
- pull to refresh in web view
- web header on notification page
  - similar look to main frame header when Republik header
  - close icon in native app
- fix piwik tracking in app

### DIFF
--- a/components/Frame/Header.js
+++ b/components/Frame/Header.js
@@ -29,14 +29,12 @@ import {
   ZINDEX_HEADER,
   ZINDEX_HEADER_BACK,
   ZINDEX_HEADER_SECONDARY,
-  ZINDEX_HEADER_LOGO
+  ZINDEX_HEADER_LOGO,
+  LOGO_HEIGHT,
+  LOGO_WIDTH,
+  LOGO_HEIGHT_MOBILE,
+  LOGO_WIDTH_MOBILE
 } from '../constants'
-
-const LOGO_HEIGHT = 28.02
-const LOGO_WIDTH = LOGO_HEIGHT * Logo.ratio
-
-const LOGO_HEIGHT_MOBILE = 22.78
-const LOGO_WIDTH_MOBILE = LOGO_HEIGHT_MOBILE * Logo.ratio
 
 const SEARCH_BUTTON_WIDTH = 28
 
@@ -206,15 +204,18 @@ const forceRefRedraw = ref => {
   if (ref) {
     const redraw = () => {
       const display = ref.style.display
+      // offsetHeight
       ref.style.display = 'none'
       /* eslint-disable-next-line no-unused-expressions */
-      ref.offsetHeight
+      ref.offsetHeight // this force webkit to flush styles (render them)
       ref.style.display = display
     }
-    setTimeout(redraw, 33)
-    setTimeout(redraw, 33 * 10)
-    setTimeout(redraw, 33 * 20)
-    setTimeout(redraw, 33 * 30)
+    const msPerFrame = 1000 / 30 // assuming 30 fps
+    const frames = [1, 10, 20, 30]
+    // force a redraw on frame x after initial dom mount
+    frames.forEach(frame => {
+      setTimeout(redraw, msPerFrame * frame)
+    })
   }
 }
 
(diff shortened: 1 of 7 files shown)


components/Frame/Header.js
```diff
@@ -203,13 +204,17 @@ const isPositionStickySupported = () => {
...
-    setTimeout(() => {
+    const redraw = () => {
       const display = ref.style.display
       ref.style.display = 'none'
       /* eslint-disable-next-line no-unused-expressions */
       ref.offsetHeight
       ref.style.display = display
-    }, 300)
+    }
+    setTimeout(redraw, 33)
```
patrickvenetz: Comments: Would leave some comments here. Questions expected to answer: Why 33ms? Why multiple times with some backoff logic?

Logic: Maybe even wrap that into an array that get's mapped.

components/Frame/Header.js
```diff
@@ -203,13 +204,17 @@ const isPositionStickySupported = () => {
 // - iOS 11.4: header is transparent and only appears after triggering a render by scrolling down enough
 const forceRefRedraw = ref => {
   if (ref) {
-    setTimeout(() => {
+    const redraw = () => {
       const display = ref.style.display
       ref.style.display = 'none'
       /* eslint-disable-next-line no-unused-expressions */
       ref.offsetHeight
```
patrickvenetz: Off PR, but out of curiosity: What good does this do?

components/Frame/Pullable.js
```diff
@@ -0,0 +1,177 @@
...
+            : <DownIcon size={32} />}
+        </Container>
+        {this.props.children}
+      </Fragment>
+    )
+  }
+}
+
+Pullable.defaultProps = {
+  distThreshold: 72,
```
patrickvenetz: Comment: Again wondering where these numbers come from, and if they could instead be derived from something? Is it like your best judgment or gut feeling?

components/Frame/Track.js
```diff
@@ -0,0 +1,39 @@
...
+    me
+      ? [].concat(me.roles).sort().join(' ') || 'none'
+      : 'guest'
+  ])
+
+// ToDo: move client side navigation tracking from LoadingBar here after next upgrade (multiple router event listener)
+
+class Track extends Component {
+  componentDidMount () {
+    const { me } = this.props
```
patrickvenetz: Code Sanity: Since used once here, would probably just go for

```
trackRoles(this.props.me)
```

pages/notifications.js
```diff
@@ -16,41 +16,83 @@ import Me from '../components/Auth/Me'
...
 import {
   CURTAIN_MESSAGE, CDN_FRONTEND_BASE_URL
 } from '../lib/constants'
 
 import {
-  Interaction, NarrowContainer, Logo, linkRule, RawHtml, mediaQueries, Button
+  Interaction, NarrowContainer, Logo, linkRule, RawHtml, mediaQueries, colors, Button
 } from '@project-r/styleguide'
 
+const LOGO_HEIGHT = 28.02
```
patrickvenetz: Structure: Might this and `LOGO_HEIGHT_MOBILE` best be placed into a "constants" file?

pages/notifications.js
```diff
@@ -136,33 +178,46 @@ const Page = withT(({ url: { query, query: { context, token, tokenType, noAutoAu
...
-            {logo}
-          </div>
-        )}
-        <div {...(inNativeApp ? styles.textApp : styles.text)}>
+        <div {...(stickyBar ? styles.bar : undefined)}>
+          {logo}
+        </div>
+        {inNativeApp && <Link route='index'>
+          <a {...styles.close}>
+            <MdClose size={32} fill='#000' />
```
patrickvenetz: Structure: May and shall this color come from SG_* ENV variable?

components/Frame/Header.js
tpreusse: It forced webkit to flush styles (render) in order to calculate `offsetHeight`.
https://stackoverflow.com/a/3485654 – I'll add a code comment

pages/notifications.js
tpreusse: Pure black is always hard coded. Never found a nice abstraction for this. Having a variable `black` which may or may not contain pure black is odd. It's the default fill of the logo and most icons. In this case leaving it out would probably also render black (default).

PR #180, opened 2018-09-01
Reload on Sign In-n-Out

tpreusse:
This also improves the app user experience: immediately reloads after most sign ins / outs instead of waiting until next navigation.


tpreusse:
Tested purchase process on staging, including claim pledge, claim membership and purchase via PayPal.

PR #182, opened 2018-09-04
Redirect /feuilleton to index page if user is not logged in

patrickvenetz:
Redirect /feuilleton to index page if user is not logged in. In case we recognize an iOS Republik-App, UnauthorizedPage is render.

Flow implemented is as follows:
1. If user is a member, render `Front` component.
2. If path is known (but user is not a member)...
    1. ... render `UnauthorizedPage` component if we recognize an iOS Republik-App
    2. ... render `Loader` component while redirect is pushed to router and requires a few moments to kick in.
3. If path is neither known (nor is user a member), render a 404 Not Found status page.

Caveats:
- `Loader` component there to avoid a moment for having render a 404 Not Found status page while router did not complete redirect yet.
- Not sure if lifecylces `componentDidMount` and `componentDidUpdate` indeed need and would.

This Pull Request is based on [redirect functionality in pages/signin.js]( https://github.com/orbiting/republik-frontend/blob/master/pages/signin.js#L16-L33)

### DIFF
--- a/pages/front.js
+++ b/pages/front.js
@@ -1,31 +1,85 @@
+import { Component } from 'react'
 import { compose } from 'react-apollo'
+
+import { Loader } from '@project-r/styleguide'
+
+import { Router, routes } from '../lib/routes'
 import Frame from '../components/Frame'
 import Front from '../components/Front'
+import StatusError from '../components/StatusError'
+
 import withData from '../lib/apollo/withData'
 import withMembership, { UnauthorizedPage } from '../components/Auth/withMembership'
-import StatusError from '../components/StatusError'
+import withInNativeApp from '../lib/withInNativeApp'
 
 const KNOWN_PATHS = ['/feuilleton']
 
-const FrontPage = (props) => {
-  if (props.isMember) {
-    return <Front {...props} />
+const isPathKnown = (url) => {
+  return KNOWN_PATHS.indexOf(url.asPath.split('?')[0]) !== -1
+}
+
+class FrontPage extends Component {
+  componentDidMount () {
+    this.redirectUser()
   }
-  const { url } = props
-  if (KNOWN_PATHS.indexOf(url.asPath.split('?')[0]) !== -1) {
-    return <UnauthorizedPage {...props} />
+
+  componentDidUpdate () {
+    this.redirectUser()
   }
-  return (
-    <Frame raw url={url}>
-      <StatusError
-        url={url}
-        statusCode={404}
-        serverContext={props.serverContext} />
-    </Frame>
-  )
+
+  redirectUser () {
+    const { url, isMember, inNativeIOSApp, serverContext } = this.props
+
+    if (isPathKnown(url) && !isMember && !inNativeIOSApp) {
+      if (serverContext) {
+        const indexPath = routes
+          .filter(r => r.name === 'index')
+          .shift()
+          .toPath()
+
+        serverContext.res.redirect(302, indexPath)
+        serverContext.res.end()
+      } else {
+        Router.pushRoute('index')
+      }
+    }
+  }
+
+  render () {
+    const { url, isMember, inNativeIOSApp, serverContext } = this.props
+
+    if (isMember) {
+      return <Front {...this.props} />
+    }
+
+    if (isPathKnown(url)) {
+      if (inNativeIOSApp) {
+        return <UnauthorizedPage {...this.props} />
+      }
+
+      // ... render Loader while redirect action is pushed to Router
+      return (
+        <Frame raw url={url}>
+          <Loader loading style={{minHeight: 'calc(100vh - 80px)'}} />
+        </Frame>
+      )
+    }
+
+    // If path is neither known (nor is user a member), render a 404 Not Found
+    // status page.
+    return (
+      <Frame raw url={url}>
+        <StatusError
+          url={url}
+          statusCode={404}
+          serverContext={serverContext} />
+      </Frame>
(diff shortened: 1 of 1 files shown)


pages/front.js
```diff
@@ -1,31 +1,83 @@
...
-  const { url } = props
-  if (KNOWN_PATHS.indexOf(url.asPath.split('?')[0]) !== -1) {
-    return <UnauthorizedPage {...props} />
+
+  redirectUser () {
+    const { url, isMember } = this.props
+
+    // If path is known but user is not a member, redirect to index page.
+    if (isPathKnown(url) && !isMember) {
+      Router.pushRoute('index')
```
tpreusse: use `replaceRoute`, don't do it when `inNativeIOSApp`

pages/front.js
```diff
@@ -1,31 +1,83 @@
...
+
+    // If user is a member, render Front.
+    if (isMember) {
+      return <Front {...this.props} />
+
+    // If path is known (but user is not a member)...
+    } else if (isPathKnown(url)) {
+      // ... display UnauthroizedPage if we recognize an iOS Republik-App
+      if (inNativeIOSApp) {
+        return <UnauthorizedPage me={me} url={url} />
```
tpreusse: IMO forward all props (spread `this.props`) and get ride of `withMe`/`me`. `UnauthorizedPage` is a page and should get all page props.

pages/front.js
```diff
@@ -1,31 +1,83 @@
...
+
+  render () {
+    const { url, me, isMember, inNativeIOSApp, serverContext } = this.props
+
+    // If user is a member, render Front.
+    if (isMember) {
+      return <Front {...this.props} />
+
+    // If path is known (but user is not a member)...
+    } else if (isPathKnown(url)) {
```
uxengine: Since you return early above, I would get rid of the else-branch and carry on with 
`if (isPathKnown(url)) {`

pages/front.js
```diff
@@ -1,31 +1,83 @@
...
+      // ... display UnauthroizedPage if we recognize an iOS Republik-App
```
uxengine: Typo: UnauthroizedPage

pages/front.js
```diff
@@ -1,31 +1,83 @@
...
+    // If path is known but user is not a member, redirect to index page.
+    if (isPathKnown(url) && !isMember) {
+      Router.pushRoute('index')
+    }
+  }
+
+  render () {
+    const { url, me, isMember, inNativeIOSApp, serverContext } = this.props
+
+    // If user is a member, render Front.
```
uxengine: I think the comment just repeats the code (which is good sign code-wise), so you could just get rid of the comment.

pages/front.js
uxengine: Again, I think the comment here mostly repeats the code.

pages/front.js
```diff
@@ -1,31 +1,83 @@
...
+      return <Front {...this.props} />
+
+    // If path is known (but user is not a member)...
+    } else if (isPathKnown(url)) {
+      // ... display UnauthroizedPage if we recognize an iOS Republik-App
+      if (inNativeIOSApp) {
+        return <UnauthorizedPage me={me} url={url} />
+      }
+
+      // ... render Loader while redirect action is pushed to router and
```
uxengine: While I think this comment is good to have.

patrickvenetz:
> We could also do server-side redirects, an example: https://github.com/orbiting/republik-frontend/blob/master/components/StatusError/index.js#L65

Loved it. Force pushed a new, shiny attempt using server-side redirects only and as far as I can tell it works like a breeze. Client-side redirect seemed no longer necessary. Please let me know if that is not the case.

> Open question: are there currently ad campaigns promoting direct links to /feuilleton?

No. There is a Medienmitteilung and [some that adopted links in it](https://www.persoenlich.com/medien/wo-kultur-besprochen-und-debattiert-wird).

pages/front.js
```diff
@@ -1,31 +1,60 @@
...
+    }
+
+    if (isPathKnown(url)) {
+      if (serverContext && !inNativeIOSApp) {
+        const indexPath = routes
+          .filter(r => r.name === 'index')
+          .shift()
+          .toPath()
+
+        serverContext.res.redirect(302, indexPath)
```
tpreusse: `/` would do too ;)

pages/front.js
```diff
@@ -1,31 +1,60 @@
...
+class FrontPage extends Component {
+  render () {
+    const { url, isMember, inNativeIOSApp, serverContext } = this.props
+
+    if (isMember) {
+      return <Front {...this.props} />
+    }
+
+    if (isPathKnown(url)) {
+      if (serverContext && !inNativeIOSApp) {
```
tpreusse: you should still handle the client case:
https://github.com/orbiting/republik-frontend/blob/3e13bd2d774b7d13e3c33fa49da6c0ba700d85ff/components/StatusError/index.js#L73
_client side links could still arrive without `serverContext`, while there shouldn't be any client side links for non members it may still happen and rendering a 404 for `/feuilleton` should always be prevented_

tpreusse: And contrary to previous review: use `pushRoute` instead of `replaceRoute`. Because one could be not logged in, arrive at `/feuilleton`, be redirect to `/`, login and want to press back to get to `/feuilleton`.

patrickvenetz:
Revised. Using `serverContext` when available, otherwise `Router.pushRoute`.

PR #183, opened 2018-09-05
Query sometimes for firstName, lastName before authorizing a token

patrickvenetz:
This Pull Request depends on https://github.com/orbiting/backends/pull/91

It does render in `Auth/TokenAuthorization` component additional fields a user has to queried for before a session can be authorized.

Information about necessary fields is provided as strings on query `unauthorizedSession` and is later submitted with mutation `authorizeSession`.

bildschirmfoto 2018-09-05 um 14 01 39


lib/translations.json
```diff
@@ -3177,6 +3177,26 @@
     {
       "key": "styleguide/DynamicComponent/error",
       "value": "Diese Komponente kann, aufgrund eines technischen Fehlers, nicht dargestellt werden."
+    },
+    {
+      "key": "pledge/fields/explanation",
+      "value": "Fast geschafft. Ihr Zugang ist ein Konto erster Güte, und Sie können unter anderem an Debatten teilnehmen. Um Ihr Profil für den Start zu vervollständigen, benötigen wir noch folgende Informationen von Ihnen:"
```
tpreusse: Use `tokenAuthorization` name space.

I would avoid wording like «Fast geschafft.», sign in is only the start :)

I'd go with some more confident and assertive:
```
Um Ihr Konto zu aktivieren, benötigen wir folgende Informationen von Ihnen:
```

lib/translations.json
```diff
@@ -3177,6 +3177,26 @@
...
+    },
+    {
+      "key": "pledge/fields/label/firstName",
```
tpreusse: Use `tokenAuthorization` name space or use existing keys.

lukasbuenger:
LGTM!

lukasbuenger:
The only thing I don't like too much, is the «generic abstractions are always better» approach to the `Fields` thing.
It really just DRYs up two (!) cases with probably as much code as would be needed to implement the fields specifically. Also, the way it's implemented right now, we'd have to extend the generic thing as soon as a new `field` value digresses even mildly from the assumpted "text with text input case" (think birthday).
You'd simply pass the following decisions to the next person:
- Should I extend the abstraction?
- If yes, isn't an abstraction, that is partially informed by the backend but still has to know literally everything about data type and validation mechanics, very incomplete? Can I make it "right" without effing up my schedule?
- If I decided to extend the abstraction, how many other clients use the same abstraction? Can I kill those clients with my update? I'd probably have to test those too, even if not related to my current job at all, in our case that means manual testing and speaking of schedules ...

Just my 2 cents.

tpreusse:
Re Fields abstraction:
I mostly think you should have used the [existing abstraction from the styleguide](https://styleguide.republik.ch/forms#field-sets) instead of creating another one ;)

I'll push two commits to resolve conflicts and migrate to the existing abstraction. Then you can decide to revert or go with it.

patrickvenetz:
@tpreusse Thanks for that grand abstraction with `FieldSet`. Scheduling for testing and release later today (backend lacks some improvements).

PR #184, opened 2018-09-05
allow to only submit name when not a customer

tpreusse:
Had to re-indent the form. Essentially this change allows someone without a membership (but e.g. with member role) to just set a name without providing an address. UX is not ideal but it works—quick fix for now.


patrickvenetz:
I assume that API does not restrict.

PR #187, opened 2018-09-11
feat(security): wear helmet

tpreusse:
Introduce all easy to implement security headers (skip CSP and HPKP for now).

HSTS initially just with a 24h window to test and ensure that no subdomain services breaks.


tpreusse:
Tested on staging, incl. dynamic component and «die Schweiz spricht»

PR #191, opened 2018-09-29
Shorter disclosure label and explanation below input field

patrickvenetz:
On small screen devices, disclosure input label did not fit onto a single. This Pull Request splits label into a label and an explanation below input field.

**Before**
vote iphone 5 label empty

**Afterwards**
vote iphone 5 new empty

### DIFF
--- a/components/Vote/ElectionCandidacy.js
+++ b/components/Vote/ElectionCandidacy.js
@@ -2,7 +2,7 @@ import React, { Fragment } from 'react'
 import ErrorMessage from '../ErrorMessage'
 import voteT from './voteT'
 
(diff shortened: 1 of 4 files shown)


components/Vote/ElectionCandidacy.js
```diff
@@ -110,6 +110,7 @@ const fields = (t, vt) => ([
   },
   {
     label: t('profile/disclosures/label'),
+    explanation: t('profile/disclosures/explanation'),
```
tpreusse: Probably should be wrapped in `<Label>`: `<Label>{t('profile/disclosures/explanation')}</Label>` (The reason not all explanation get auto wrapped is flexibility—any react node can be passed—strings are valid too but probably we want the small grey type here)

patrickvenetz: It should and it will.

PR #193, opened 2018-10-01
Wahlplattform Bugfixes

garamond:


components/Vote/ElectionCandidacy.js
```diff
@@ -216,11 +218,12 @@ class ElectionCandidacy extends React.Component {
   }
 
   deriveStateFromProps ({data}) {
-    const {statement, birthday, disclosures, credentials, address, portrait} = data.me || {}
+    const {username, statement, birthday, disclosures, credentials, address, portrait} = data.me || {}
     const {line1, line2, city, postalCode, country = DEFAULT_COUNTRY} = address || {}
     const credential = credentials ? credentials.find(c => c.isListed) : {}
     return {
       values: {
+        username: username ? undefined : username,
```
tpreusse: Couldn't this line be removed completely? and in place add a comment that we do not pre-fill username in order to not overwrite it (when no field is shown)? Alternatively add a comment explaining what this line does and why.

garamond: Very true

PR #200, opened 2018-10-15
discussion: misc improvements

uxengine:
Related to
https://github.com/orbiting/backends/pull/118
https://github.com/orbiting/styleguide/pull/181

Changelog:
- Default sort option `newest` instead of `hottest`
- Drop `hottest` sort option in favor of "most answers"
- collapsable comments
- upVotes/downVotes instead of score
- share button for comments


uxengine:
@tpreusse 
Focus-specific meta tags are implemented now as discussed. I decided not to change the generic meta description tag to the comment's content, but only the twitter/facebook description tags. The idea of someone random being able to influence our pages' description tags didn't sound right to me.

screen shot 2018-10-24 at 12 28 02

uxengine:
@tpreusse Actually, the server-side page source doesn't have the meta tags (now that they're rendered outside frame). Need your advise ...

components/Article/Page.js
```diff
@@ -411,7 +411,8 @@ class ArticlePage extends Component {
     return (
       <Frame
         raw
-        meta={meta}
+        // Meta tags for a focus comment are rendered in Discussion/Commments.js
+        meta={router.query.focus ? undefined : meta}
```
tpreusse: check that `meta.discussionId` is present too, otherwise one could disable meta data on a non discussion by setting focus param that has no other effect

components/Discussion/Comments.js
```diff
@@ -586,12 +615,35 @@ class Comments extends PureComponent {
             />
           )
 
+          const shareOverlay = !!shareUrl && (
+            <ShareOverlay
+              discussionId={discussionId}
+              onClose={() => this.toggleShare(null)}
```
tpreusse:
```suggestion
              onClose={() => this.setState({ shareUrl: undefined })}
```

components/Discussion/Comments.js
```diff
@@ -465,6 +490,8 @@ class Comments extends PureComponent {
           replyBlockedMsg={replyBlockedMsg}
           Link={CommentLink}
           secondaryActions={<SecondaryActions />}
+          collapsable={discussion && discussion.collapsable}
+          onShare={() => { this.toggleShare(discussion.documentPath, comment.id) }}
```
tpreusse:
```suggestion
          onShare={() => this.setState({ shareUrl: getFocusUrl(discussion.documentPath, comment.id) }) }}
```

components/Discussion/Comments.js
```diff
@@ -83,6 +99,14 @@ class Comments extends PureComponent {
           }
         })
     }
+
+    this.toggleShare = (path, commentId) => {
```
tpreusse: since you're not really toggling here I'd just use `setState` directly, see suggestion

PR #204, opened 2018-10-17
Wahlplattform Hotfix "The Morning After"

garamond:


components/Vote/text.js
```diff
@@ -44,22 +45,32 @@ const PMedium = (props) =>
...
-        .map((c, i) =>
-          <PMedium key={i}>
-            <RawHtml dangerouslySetInnerHTML={{ __html: c }} />
-          </PMedium>
-        )
-    }
-  </div>
+export const Body = withInNativeApp(({ dangerousHTML, inNativeApp }) => {
+  const html = inNativeApp
+    ? dangerousHTML.replace(/'/g, '"').replace(/target="_blank" /, '')
```
tpreusse: RE `target="_blank" ` replace:
Needs `g` flag too, why with an space at the end? `<a href="x" target="_blank">` is valid too

components/Vote/text.js
```diff
@@ -78,17 +89,23 @@ const PSmall = ({ children, indent = true }) =>
...
-        .map((c, i) =>
-          <PSmall key={i} indent={indent}>
-            <RawHtml dangerouslySetInnerHTML={{ __html: c }} />
-          </PSmall>
-        )
-    }
-  </div>
+export const Small = withInNativeApp(({ dangerousHTML, inNativeApp, indent = true }) => {
+  const html = inNativeApp
+    ? dangerousHTML.replace(/'/g, '"').replace(/target="_blank" /, '')
```
tpreusse: see above, maybe move to util fn

components/Vote/text.js
garamond:
```suggestion
    ? dangerousHTML.replace(/'/g, '"').replace(/target="_blank"/g, '')
```

components/Vote/text.js
garamond:
```suggestion
    ? dangerousHTML.replace(/'/g, '"').replace(/target="_blank"/g, '')
```

PR #209, opened 2018-10-18
Small improvements

garamond:


components/Vote/ElectionBallotRow.js
```diff
@@ -224,8 +224,8 @@ class ElectionBallotRow extends Component {
             {
               summary
             }
-            <div>
-              { showMeta &&
+            { showMeta &&
+              <div>
```
tpreusse: Ideally we'd wouldn't use child selectors in this cases. And assign a css to the summary `div`s here, now it's not very clear why this divs is actually needed.

PR #238, opened 2018-12-20
Reading progress

uxengine:
Depends on https://github.com/orbiting/styleguide/pull/206

## Life-cycle in a nutshell:
- new higher order component `Progress` for managing reading progress, which receives these props from `Article`:
  - `article`
  - `isMember`
- `Progress` calls `initializeProgress(userProgress, embeds)` once (on mount or update)
- if there's no user consent yet, `Progress` renders a `ProgressPrompt`
- if user rejects, nothing will be tracked.
- if there is a user consent, `Progress` ...
  - shows a Spinner
  - polls the DOM until it finds elements tagged with `data-pos`
  - restores the persisted `userProgress` if any, preferrably from `nodeId` or alternatively from `percentage`
  - provides `getMediaProgress` and `saveMediaProgress` callbacks via `context` to the media players.
  - hides the spinner
- once `userProgress` is restored (if any), `Progress` is initialized and ready to save progress
- `Progress` calls `saveProgress` on scroll (debounced to scroll end)
- `Progress` detects the first visible article in viewport and persists `nodeId` and `percentage`. Note that it internally keeps track of the element index and traverses the element list downwards or upwards depending on scroll.
- the progress of media elements (only our own Vimeo videos and audio files) is captured in two ways:
  - throttled: while playing, the current media time is upserted every (5) seconds
  - debounced: on pause, video end or scrubbing around

For re-enabling the progress prompt locally while testing:
`DELETE FROM consents where policy = 'PROGRESS';`

### If the user hasn't confirmed or rejected the consent yet, a prompt appears
screenshot 2019-02-12 at 15 49 06

### After confirmation, the reading position is tracked for a specific document while scrolling (but only persisted when scrolling down):
screenshot 2019-02-12 at 15 49 44

### After navigating away and coming back to the same article, the reading position is restored both on desktop ...
screenshot 2019-02-12 at 15 50 30

### ... and mobile
screenshot 2019-02-12 at 15 53 03

### Consent can be revoked or granted on the account page
screenshot 2019-02-12 at 15 53 49

### Feed icon indicates progress
screenshot 2019-02-12 at 16 04 14


components/Article/withReadingProgress.js
```diff
@@ -220,14 +220,14 @@ const withReadingProgress = WrappedComponent => {
             progressElementIndex: nextIndex
           })
           return {
-            nodeId: progressElement && progressElement.id,
+            nodeId: progressElement && progressElement.dataset && progressElement.dataset.pos,
```
tpreusse: use `getAttribute('data-pos')` instead, needs no extra check for `dataset` presence

uxengine:
@tpreusse This is ready for a first round of comments.  I've added some notes about the life cycle.`ProgressPrompt` is WIP, but it already provides the core interactions (confirming or rejecting the `PROGRESS` consent).

uxengine:
@tpreusse FYI, the videoplayer currently gets remounted for some reason after scrolling down (i.e. after upsertDocumentProgress mutation fires). That causes the player to re-initialize and the startTime to jump back to 0 for a moment after each downwards scroll. Don't exactly understand why yet ...

uxengine:
@tpreusse Thanks, both suggestions are addressed now.

components/Article/Progress/TopButton.js
```diff
@@ -0,0 +1,78 @@
...
+import UpIcon from 'react-icons/lib/md/arrow-upward'
+import { negativeColors } from '../../Frame/constants'
+
+const ANIMATE_OUT_DURATION_SECS = 0.3
+const RADIUS = 16
+
+const styles = {
+  container: css({
+    backgroundColor: negativeColors.primaryBg,
+    boxShadow: '0 0 0px 2px rgba(255, 255, 255, .25)',
```
garamond: Nice touch!

components/Article/Progress/TopButton.js
```diff
@@ -0,0 +1,78 @@
...
+    width: `${RADIUS * 2}px`,
+    height: `${RADIUS * 2}px`,
+    borderRadius: `${RADIUS}px`,
+    border: 'none',
+    outline: 'none',
+    transition: `opacity ${ANIMATE_OUT_DURATION_SECS}s ease-in-out`
+  })
+}
+
+class TopButton extends React.Component {
```
garamond: Isn't this a bit of a misnomer? After all, the button is nowhere near the top…

components/Article/Progress/index.js
```diff
@@ -0,0 +1,427 @@
...
+import ProgressPrompt from './ProgressPrompt'
+import TopButton from './TopButton'
+import { Spinner, mediaQueries } from '@project-r/styleguide'
+
+import { HEADER_HEIGHT, HEADER_HEIGHT_MOBILE, ZINDEX_POPOVER } from '../../constants'
+
+import { withProgressApi } from './api'
+
+const MAX_POLL_RETRIES = 3
+const SCROLLED_AWAY_PX = 200
```
garamond: IMO this could be less

components/Article/Progress/index.js
```diff
@@ -0,0 +1,427 @@
...
+      if (this.container) {
+        const { width, height, top } = this.container.getBoundingClientRect()
+        const cleanWidth = Math.min(width, 695)
+        if (cleanWidth !== this.state.width || height !== this.state.height || top !== this.state.top) {
+          this.setState({ width: cleanWidth, height, top })
+        }
+      }
+    }
+
+    this.measureProgress = (downwards = true) => {
```
garamond: There is a lot happening in this method. Given that it is called on every scroll (debounced), we should look into some optimizations.

components/Article/Page.js
```diff
@@ -258,6 +282,10 @@ class ArticlePage extends Component {
     }
 
     this.measure = () => {
+      const mobile = window.innerWidth < mediaQueries.mBreakPoint
```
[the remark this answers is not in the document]
uxengine: Yes, because it's only called in `componentDidMount` and `componentDidUpdate` which are not called server-side.

components/Article/Progress/ProgressPrompt.js
```diff
@@ -0,0 +1,103 @@
...
+
+const styles = {
+  box: css({
+    paddingTop: 10,
+    paddingBottom: 10,
+    position: 'relative',
+    backgroundColor: colors.primaryBg,
+    [mediaQueries.mUp]: {
+      paddingTop: 30,
+      paddingBottom: 30
```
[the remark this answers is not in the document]
uxengine: This was caused by some extra unwanted margin of the last content paragraph. fixed.

components/Article/Page.js
```diff
@@ -198,8 +216,10 @@ class ArticlePage extends Component {
           }
         })
       } else {
+        const showAudioPlayer = !this.state.showAudioPlayer
         this.setState({
-          showAudioPlayer: !this.state.showAudioPlayer
+          showAudioPlayer,
+          headerAudioPlayer: showAudioPlayer ? this.getAudioPlayer() : null
```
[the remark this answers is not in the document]
uxengine: The rationale is sad but true: This is the only way I could make things run stable. If I create that `<Progress><AudioPlayer /></Progress>' in `render`, the audio player will constantly re-render and visually flicker while playing. I haven't gone to the bottom of this yet ...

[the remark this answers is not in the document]
garamond: Maybe `<AudioPlayer key='audioPlayer' />`?

PR #244, opened 2019-01-13
Communicate Deactivation and Improve Membership Explanations

patrickvenetz:
This Pull Request will ensure a cancelled membership (`renew: false`) can only be reactivated if no prolong is available. An expired membership requires a prolong to be available.

It attempts to improve "state wording" about a membership:

* "Automatische Erneuerung am {formattedEndDate}" displayed not only for `MONTHLY_ABO`
* "Gültig bis am {formattedEndDate}" on cancelled but not expired memberships
* "Endete am {formattedEndDate}" on (auto-)cancelled and expired memberships

bildschirmfoto 2019-01-13 um 14 52 40

bildschirmfoto 2019-01-13 um 14 54 26

It renders a note about no active memberships in a first attempt to improve understanding of "Abonnements" section.

bildschirmfoto 2019-01-13 um 14 49 46

### T-30 (as is)
screenshot 2019-01-14 at 00 08 14

### T-2
screenshot 2019-01-14 at 00 07 40

### T+1
screenshot 2019-01-14 at 00 04 44

### DIFF
--- a/components/Account/Memberships/Manage.js
+++ b/components/Account/Memberships/Manage.js
@@ -57,17 +57,7 @@ class Actions extends Component {
         {!prolong && membership.active && membership.renew && waitingMemberships && <P>
           {t('memberships/manage/prolong/awaiting')}
         </P>}
-        {membership.active && membership.renew && <P>
-          <Link route='cancel' params={{ membershipId: membership.id }} passHref>
-            <A>
-              {t.first([
-                `memberships/${membership.type.name}/manage/cancel/link`,
-                'memberships/manage/cancel/link'
-              ])}
-            </A>
-          </Link>
-        </P>}
-        {!membership.renew && !!membership.periods.length && membership.type.name === 'MONTHLY_ABO' && <P>
+        {!membership.renew && !!membership.periods.length && !prolong && <P>
           <A href='#reactivate' onClick={(e) => {
             e.preventDefault()
             this.setState({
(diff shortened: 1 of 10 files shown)


components/Account/Memberships/Manage.js
```diff
@@ -154,8 +154,16 @@ const ManageActions = compose(
...
+  const latestPeriod = membership.periods
+    .reduce((acc, period) => {
+      return acc && new Date(period.endDate) < new Date(acc.endDate)
+        ? acc
+        : period
+    })
+
+  const latestPeriodEndDate = latestPeriod && new Date(latestPeriod.endDate)
+  const formattedEndDate = latestPeriod && dayFormat(latestPeriodEndDate)
+  const overdue = latestPeriod && latestPeriodEndDate < new Date()
```
tpreusse: I'd prefer to keep this logic on the server: [crowdfundings/graphql/resolvers/Membership.js](https://github.com/orbiting/backends/blob/28346f15c64a00451d7668dea735b7c5c124344f/servers/republik/modules/crowdfundings/graphql/resolvers/Membership.js#L9). Maybe it can be simplified there, in the monthly case it is nice to ensure that a charge attempt was made.

components/Frame/ProlongBox.js
```diff
@@ -25,28 +29,67 @@ const styles = {
...
+    const dark = inDarkFrame || key !== 'before'
+    const colorStyle = { color: dark ? '#fff' : undefined }
+
+    const explanation = t.elements(`${baseKey}/explanation`, {
+      cancelLink: <Link key='cancelLink' route='cancel' passHref>
+        <Editorial.A style={colorStyle}>
+          {t(`${baseKey}/explanation/cancelText`)}
+        </Editorial.A>
+      </Link>,
+      graceEndDate: dayFormat(timeDay.offset(date, 14))
```
patrickvenetz: Shall we deliver that `graceEndDate` that via API to ensure backend and frontend are in sync?

PR #267, opened 2019-02-25
Progress: misc launch-blocking fixes

uxengine:
Misc fixes:
- don't restore document position at 100%
- display max percentage in teasers if it's 100%
- enable progress circle on article teasers also on Bookmarks page
- add more user guidance and visibility to back-to-top button
- load mediaProgress independent of document (so that audio progress refreshes cross-device on close and re-open)

Depends on https://github.com/orbiting/styleguide/pull/219

### DIFF
--- a/components/Article/Progress/index.js
+++ b/components/Article/Progress/index.js
@@ -244,12 +244,14 @@ class Progress extends Component {
           // We only persist progress for a downward scroll, but we still measure
           // an upward scroll to keep track of the current reading position.
           const progress = this.measureProgress(downwards)
+          const storedUserProgress = this.props.article && this.props.article.userProgress
           if (
             downwards &&
             progress &&
             progress.nodeId &&
             progress.percentage > 0 &&
-            progress.elementIndex > 1 // ignore first two elements.
+            progress.elementIndex > 1 && // ignore first two elements.
+            (!storedUserProgress || storedUserProgress.nodeId !== progress.nodeId)
           ) {
             this.props.upsertDocumentProgress(documentId, progress.percentage, progress.nodeId)
           }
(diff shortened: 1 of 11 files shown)


lib/translations.json
```diff
@@ -3048,7 +3048,23 @@
...
+      "key": "article/progress/time/today",
+      "value": "von heute"
+    },
+    {
+      "key": "article/progress/time/yesterday",
+      "value": "von gestern"
+    },
+    {
+      "key": "article/progress/time/other",
+      "value": "vom {date}"
```
patrickvenetz:
```suggestion
      "value": "am {date}"
```

Schweizer nutzen "*am Tag.Monat.Jahr"

uxengine: Merci vielmal. Isch jetz helvetisiert, au de Tescht.

components/Article/Progress/index.js
```diff
@@ -333,8 +321,19 @@ class Progress extends Component {
     }, 5000, { 'trailing': false })
 
     this.getMediaProgress = (mediaId) => {
-      const { mediaProgress } = this.state
-      return mediaProgress ? mediaProgress[mediaId] : undefined
+      return new Promise((resolve, reject) => {
```
tpreusse: directly return `this.props.client.query().then()`, `this.props.client.query` already returns a promise, and `then` on a promise also always returns another promise.

No `catch` here, it's fine to swallow the error in the audio / video player but IMO no need to already swallow here. Also in the current implementation if `data.mediaProgress || data.mediaProgress === null` does not pass the promise is never resolved or rejected.

tpreusse: API docs on `errors` is not super clear, at least I didn't find it. Therefore just ignore, I guess. In theory the promise should also be rejected.

PR #269, opened 2019-02-26
Propagate query params of recognized hostname links in Link/Href

patrickvenetz:
This Pull Request will parse query parameters passed via `href` variable and merge with `query` variable provided. Latter super-seeds URL query parameters.


patrickvenetz:
Thanks.

Sure. I mean... everything is staged first, isn't it... 🙈

PR #271, opened 2019-02-28
Disable new submissions on statements

garamond:
This fix is very superficial, pending modularization of the fuctioniality


patte:
closed because I extracted everything into a DynamicComponent :tm: 
see: https://github.com/orbiting/crowd-questions

Issue #272, opened 2019-03-04
"Abo teilen" button redirects to marketing page after email confirmed.


eightam:
## Bug Report

If an "Abo teilen" is accepted and the confirmation link is opened in the same browser a page with a big green button "Magazin lesen" appears. If the button is clicked, the page jumps to the marketing front. If the browser is refreshed, the user is logged in and on the magazine front.

**Steps to Reproduce**

Get an Abo shared to you, accept the invitation, log in in the same browser, click the button.

**Actual Results**

If an "Abo teilen" is accepted and the confirmation link is opened in the same browser a page with a big green button "Magazin lesen" appears. If the button is clicked, the page jumps to the marketing front. If the browser is refreshed, the user is logged in and on the magazine front.

**Expected Results**

The "Magazin lesen" button should jump to the magazine front.

**Environment**

Tested on Desktop Browsers
- OS: OSX 10.14.3 and Ubuntu
- Browser: Chrome 71, Firefox 65.01

**Possible Solutions**

The "Magazin lesen" button should do a hard refresh and jump to the magazine front.

patrickvenetz:
Unable to reproduce in staging environment. «Magazin lesen» does not appear after sign in completed.

Can be reproduced in production environment.

PR #279, opened 2019-03-12
Changes to Access Grant flow

patrickvenetz:
This Pull Request addresses these issues:
* Redirect from claim page to index page after a successful claim (instead of account page)
* Button on notification page after successful login when claiming an access grant won't be shown (as it isn't on preview and membership claims)

Fixes https://github.com/orbiting/republik-frontend/issues/272

### DIFF
--- a/components/Account/Memberships/Claim.js
+++ b/components/Account/Memberships/Claim.js
@@ -172,23 +172,22 @@ class ClaimMembership extends Component {
     const claim = () => {
       const code = sanitizeVoucherCode(values.voucherCode)
 
-      const claimWith = (fn, { code, context = 'unknown' }) =>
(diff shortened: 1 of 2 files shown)


components/Account/Memberships/Claim.js
```diff
@@ -165,20 +172,26 @@ class ClaimMembership extends Component {
     const claim = () => {
       const code = sanitizeVoucherCode(values.voucherCode)
 
-      if (isAccessGrantVoucherCode(code)) {
-        this.props.claimAccess(code)
+      const claimWith = (fn, { code, context = 'unknown' }) =>
```
tpreusse: `fn` -> `mutate` oder `mutation`

PR #295, opened 2019-04-26
Hide PGP link in Profil when in app

patrickvenetz:


components/Profile/Contact.js
```diff
@@ -202,7 +204,7 @@ const Contact = ({ user, isEditing, onChange, values, errors, dirty, t }) => {
           />
         )}
       </div>
-      {user.pgpPublicKeyId &&
+      {!inNativeIOSApp && user.pgpPublicKeyId &&
```
tpreusse: did you test that it doesn't cause the same issue on Android?

patrickvenetz: No, I did not will. Do so shortly.

patrickvenetz: Checked on two mobile Android devices, neither of them is acting on that link.

tpreusse: IMO the right approach would then be to either hide it in on both platforms or remove the link on both plattform and still show the key id (still—somehwat—useful info).

PR #309, opened 2019-05-24
Dialog 4.0

wereHamster:
This PR integrates to updated Discussion components into the frontend.

* All related graphql queries and mutations have been moved into  `components/Discussion/graphql/`
* The large and complicated …/Discussion/Components.js file is now much simpler. Pretty much all it does is wait until everything is loaded, assemble the comment tree (out of the flat list that is stored in the Apollo cache), create the value for the DiscussionContext and use `<CommentList>` to render the tree starting at the root comments.


components/Discussion/Comments.js
```diff
@@ -233,470 +234,202 @@ class Comments extends PureComponent {
...
+                } else {
+                  /* XXX: When does this happen? */
+                  return <React.Fragment children={props.children} />
+                }
+              }
+            },
+            composerSecondaryActions: <SecondaryActions />
+          }
+
+          const OrderBy = ({ children, value }) => (
```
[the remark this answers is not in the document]
wereHamster: I'm surprised standardjs doesn't issue a warning (unused variable, or unused parameter) when `children` is unused.

wereHamster:
Regarding fading out headers of collapsed comments: We'd need to go as low as 0.5 opacity, and even then difference between a normal header and a faded out header is barely visible. If we go even lower then the text becomes unreadable due to low contrast. For that reason Chris suggested to not do this. I can make the change in a separate commit and you can decide for yourself.

wereHamster:
What do you mean with "show focus error before comment tree somewhere"?

tpreusse:
@wereHamster the `focusError` state of `components/Discussion/Comments.js` is never displayed. IMO should be displayed between root composer and comment tree.

_Probably pre-existing condition._

wereHamster:
See the error I opened in the backend repo. I can maybe work around it in the front end though.

tpreusse:
@wereHamster I see. IMO just display that error and we will take care of fixing the backend.

components/Discussion/graphql/store.js
```diff
@@ -88,13 +88,11 @@ export const mergeComments = ({ parentId, appendAfter, comments }) => draft => {
 
     /*
      * Update the 'CommentConnection' of the parent comment.
-     *
-     * We clear 'nodes' because we don't need it. Though we shouldn't request it
-     * in the query at all, it's likely a mistake if we requested it.
      */
     draft.discussion.comments.nodes[parentIndex].comments = {
-      ...comments,
```
tpreusse: `comments` is a `CommentConnection` which needs an `id` and `__typename`. IMO the old spread solution was IMO fine.

wereHamster: I reverted the commit, and removed the comment above. GraphQL appears to do the right thing already.

PR #311, opened 2019-05-31
Remove list of paths on which to keep passed query

patrickvenetz:
Depends on https://github.com/orbiting/backends/pull/270

```gql
query { redirection(path:"/some-path?name=value") { target } }
```

If redirection is configured properly, `target` will overwrite query parameters passed in `path` and return in `target`.

Thus far they were stripped, and only kept for a dedicated set of URLs.

### DIFF
--- a/components/StatusError/index.js
+++ b/components/StatusError/index.js
@@ -48,7 +48,7 @@ export default compose(
     skip: props => props.statusCode !== 404 || !props.router.asPath,
     options: ({ router: { asPath } }) => ({
       variables: {
-        path: asPath
+        path: asPath.split('#')[0]
       }
     }),
     props: ({ data, ownProps: { serverContext, statusCode, router, inNativeApp, me } }) => {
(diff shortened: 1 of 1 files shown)


components/StatusError/index.js
```diff
@@ -54,7 +48,7 @@ export default compose(
     skip: props => props.statusCode !== 404 || !props.router.asPath,
     options: ({ router: { asPath } }) => ({
       variables: {
-        path: cleanAsPath(asPath)
+        path: asPath
```
tpreusse:
```suggestion
        path: asPath.split('#')[0]
```

tpreusse: > Clients are not supposed to send URI-fragments to servers when they retrieve a document, and without help from a local application (see below) fragments do not participate in HTTP redirections.
[Fragment identifier on Wikipedia](https://en.wikipedia.org/wiki/Fragment_identifier)

I'd follow that and also avoid sending fragment up here.

PR #315, opened 2019-06-13
Sort upcoming and past events in Events/List

patrickvenetz:
This Pull Request sorts upcoming events in an ascending («next event first»), past events in a descending one («most recent first»).

Before past events listed oldest first.

### DIFF
--- a/components/Events/List.js
+++ b/components/Events/List.js
@@ -54,9 +54,7 @@ const Overview = compose(
   graphql(query, {
     props: ({ data, ownProps: { router: { query: { slug } }, t } }) => {
       const error = data.error
-      const events =
-        data.events &&
-        data.events.map(event => ({ ...event, __parsedDate: parseDate(event.date) }))
+      const events = data.events.map(event => ({ ...event, __parsedDate: parseDate(event.date) }))
       let event
 
       if (slug && events && !error) {
--- a/components/Events/List.js
+++ b/components/Events/List.js
@@ -1,26 +1,25 @@
 import React from 'react'
 import { compose, graphql } from 'react-apollo'
 import gql from 'graphql-tag'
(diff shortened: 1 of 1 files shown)


components/Events/List.js
```diff
@@ -55,15 +54,19 @@ const Overview = compose(
...
+
+      if (slug && events && !error) {
+        event = events.find(event => event.slug === slug) || 404
       }
       return {
         loading: data.loading,
-        events: data.events,
-        error,
-        event
+        events,
```
garamond: if `data.events` is falsy `events` should be an empty array

components/Events/List.js
```diff
@@ -101,12 +104,13 @@ const Overview = compose(
...
-      })
-      const past = events.filter(event => {
-        return today > parseDate(event.date)
-      })
+      const upcoming = events
+        .filter(event => today <= event.__parsedDate)
+        .sort((a, b) => ascending(a.__parsedDate, b.__parsedDate))
+      const past = events
+        .filter(event => today > event.__parsedDate)
+        .sort((a, b) => descending(a.__parsedDate, b.__parsedDate))
```
garamond: why are past event sorted descendig?

components/Events/List.js
patrickvenetz: If not mistaken,`data.events` can't be `null` or otherwise falsy due to schema restrictions?

```
type queries {
  [...]
  events: [Event!]!
  [...]
}
```

garamond: ok, in that case the `data.events && …` test above is unnecessary

patrickvenetz: True! Adopting.

Thank you.

patrickvenetz:
> LGT… you know the rest

image

PR #316, opened 2019-06-13
Fix Discussion v4

tpreusse:
- fix meta data for focus links
- unify deep link logic for focus link
- disable share unless public profile
- fix share image for profiles
- make verified role check mark green everywhere
- use User.slug in most places


lib/timeduration.test.js
```diff
@@ -7,61 +7,61 @@ import { t } from './withT'
...
     new Date(2018, 0, 11, 0, 0, 0),
-    '1h'
+    'vor 1 h'
   ],
   [
     'days',
     new Date(2018, 0, 13, 0, 0, 0),
     new Date(2018, 0, 11, 0, 0, 0),
-    '2t'
+    'vor 2 d'
```
patrickvenetz: Would probably give these translation strings a go remove denglish, so:

"vor 2 d" becomes "vor 2 T"
"vor 1 w" becomes "vor 1 W"
"vor 1 y" becomes "vor 1 J"

While "s", "m" and "h" are :+1:

tpreusse: Dani Meyer ultimately decided for English all the way

PR #339, opened 2019-08-14
Add Trial components and replace preview routing

patrickvenetz:
This Pull Request replaces preview feature with "trail access". This is an adopted access grant flow in which a new mutation will grant and claim access (also know as "Republik teilen").

You can sign up for "trial access" on /probelesen.

a) If your logged in, it will show you a button which requests a "trial access".
b) If you're not logged in yet, it will show you an email address field and consent boxes. Form will usher you through `EMAIL_CODE` sign-in flow before requesting "trail access".

(If your logged in with an active membership or access grant, you will be redirect to front.)

Requesting a "trail access" will use [(new) mutation `requestAccess`](https://github.com/orbiting/backends/pull/297).

---

Deploy version available on https://login.republik.black/

---

This change requires **a new environment variable `TRIAL_CAMPAIGN`** with an ID of a campaign for trial accesses.

**Example trail campaign.** Grants access for 10 minutes and follows up 10 minutes later. A user can only claim one trial - ever.

Requires empty `accessGrants` table and Access Scheduler enabled (set `ACCESS_SCHEDULER=true` in `.env`)

```sql
INSERT INTO "public"."accessCampaigns"("title","description","grantPeriodInterval","emailFollowup","beginAt","endAt","constraints","grantClaimableInterval")
VALUES
(E'Test Trial Access',NULL,E'00:10:00',E'00:10:00',now(),E'2020-01-01 00:00:00+00',E'[{"recipientIsGranter": {}}, {"notVisible": {}}, {"limitGrants": {"grants": 1}}]',E'00:01:00')  RETURNING "id";
```

---

ToDos:
- [ ] Settle on text
- [ ] Show pay note on articles (e.g. `withoutActiveMembership` wrapper)

Depends on https://github.com/orbiting/backends/pull/297


components/Marketing/Trial.js
```diff
@@ -0,0 +1,211 @@
...
+
+  const errorMessages = [email.error]
+    .concat(getConsentsError(t, REQUIRED_CONSENTS, consents))
+    .filter(Boolean)
+
+  return (
+    <Fragment>
+      <H1>Danke für Ihre Neugier!</H1>
+      <P style={{ marginTop: 40 }}>
+        Bitte tragen Sie unten Ihre E-Mail-Adresse ein. Im Gegenzug laden wir Sie auf einen Streifzug
```
garamond: Hard-coded string should be moved to translations.

components/Marketing/Trial.js
```diff
@@ -0,0 +1,211 @@
...
+              ))}
+            </ul>
+          </div>
+        )}
+
+        {!signingIn && (
+          <div {...styles.button}>
+            {loading
+              ? <InlineSpinner />
+              : <Button
```
garamond: The button could be disabled until the form is valid.

components/Marketing/Trial.js
```diff
@@ -0,0 +1,211 @@
...
+  mutation requestAccess($campaignId: ID!) {
+    requestAccess(campaignId: $campaignId) {
+      id
+      endAt
+    }
+  }
+`
+
+const toOnboarding = () => {
+  window.location = format({ pathname: '/einrichten', query: { context: 'trial' } })
```
garamond: Is this necessary, wouldn't `Router.replaceRoute` suffice?

components/Marketing/Trial.js
```diff
@@ -0,0 +1,211 @@
...
+}
+
+const Trial = (props) => {
+  const { me, t } = props
+
+  const [consents, setConsents] = useState()
```
garamond: IMO, initial state should always be explicit, in this case `useState(null)` if that's what you want.

components/Marketing/Trial.js
patrickvenetz: Yes, strings ought to be replaced with translations keys once settle on text.

components/Marketing/Trial.js
patrickvenetz: Having it not disabled initially allows for a user to click it and see a list of errors what requires to be filled.

I like this pattern quite much and we're using it in other forms, too. Would like to keep it that way.

components/Marketing/Trial.js
patrickvenetz: Thus far replacing was not enough to trigger a refetch of `me` query.

However: Apollo provides a `refetch` method to each query. Did expose that on `withMe` as `meRefetch`. This is then invoked before `Router.replaceRoute` to on-boarding page. It seems to do the trick, and in a less brute-force manner.

PR #343, opened 2019-08-27
Improve share feature

annatraussnig:
add 3 small improvement on existing share functionality:

1. Change share icon to match iOS style when in iOS
1 iOS

2. Add possibility to copy link to clipboard
2 copyLink

3. Render all the share buttons at the bottom of the article
3 renderBottomPage


components/ActionBar/ShareButtons.js
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+
+  const copyLinkStatuses = {
+    init: 'init',
+    success: 'success',
+    error: 'error'
+  }
+
+  const [linkCopyStatus, setShouldLinkCopyStatus] = useState(copyLinkStatuses.init)
```
annatraussnig: remove should

PR #358, opened 2019-10-14
Slider fixes

annatraussnig:
- Include preview text in the font size overlay

Screenshot 2019-10-14 at 16 49 31


components/FontSize/Overlay.js
```diff
@@ -51,17 +77,30 @@ const FontSizeOverlay = ({ t, onClose }) => {
...
       <OverlayBody>
         <div>
-          <label {...labelStyle}>{fontPercentage.current}</label>
+          <label {...styles.label}>{fontPercentage.current}</label>
           <Slider
             value={fontSize}
             min='8'
             max='48'
-            step='4'
+            step='0.16'
```
annatraussnig: Miri mentioned that some of the feedback was a more fine grained slider – adjusting the step would be a super easy fix to get a percent-level slider. Just a suggestion.

tpreusse: Seems excessive to me. Did you get exact feedbacks? And how many people wanted more steps? Maybe reduce to `2`?

Btw. I recently learned that Miriam does not like the Nickname Miri. So we have to call her Mimi or Miriam :)

annatraussnig: Oh that was an honest mistake – I feel like I heard literally everybody call her Miri.
Shall I simply revert, compromise with `step='2'` or does it make sens to get teh numberz from Miriam when she comes in and then make a decision?

tpreusse: Let's do `0.8`

tpreusse:
Remove font size button from community & profile.

PR #360, opened 2019-10-24
Upgrade deps: React, Next & Apollo

garamond:


tpreusse:
I research the new `forwardRef` requirements for links.

It now does `IntersectionObserver`s to preload linked pages:
https://github.com/zeit/next.js/blob/a964dc825ecea1e35fb460184828974e00b7d6d6/packages/next/client/link.tsx#L132-L139

I think it's a bit much. But I guess it makes sense for mobile because mouse enter do not happen. 

My conclusion is we should add `forwardRef` to `A` and `Editorial.A` in the styleguide (and properly forward to the a tag). This should cover 90% of the case. For the rest IMO we can ignore the warning and implement it later.

PR #361, opened 2019-10-24
WIP: Replace `standard` formatter with `eslint` & Prettier 

garamond:


components/Vote/ElectionCandidacy.js
```diff
@@ -145,7 +145,7 @@ class ElectionCandidacy extends React.Component {
   constructor (props) {
     super(props)
     this.state = {
-      isEditing: props.router.query.hasOwnProperty('edit') || false,
+      isEditing: props.router.query['edit'] || false,
```
tpreusse:
```suggestion
      isEditing: !!props.router.query.edit || false,
```
should probably be cast to boolean

components/Frame/Footer.js
```diff
@@ -152,7 +152,7 @@ class Footer extends Component {
               <a
                 href='https://www.google.ch/maps/place/Sihlhallenstrasse+1,+8004+Zürich'
                 target='_blank'
-                rel='noopener'
+                rel='noopener noreferrer'
```
tpreusse: we already set the meta tag in `pages/_document.js`:
```
<meta name='referrer' content='no-referrer' />
```
And the `no-referrer` header via `herlemt` in `server/index.js`. If the linter wants this: disable the rule.

PR #363, opened 2019-11-07
WIP Paynote fixes

annatraussnig:
_TODO: campaign ID & tracking with @patrickvenetz _

### Fix buttons in mobile view
 
Screenshot 2019-11-07 at 10 25 10

### DIFF
--- a/components/Trial/Form.js
+++ b/components/Trial/Form.js
@@ -40,19 +40,15 @@ const styles = {
     width: 170,
     textAlign: 'center'
   }),
-  successActions: css({
+  buttonRow: css({
     display: 'flex',
     flexDirection: 'column',
-    [mediaQueries.mUp]: {
-      display: 'inherit'
-    }
-  }),
-  buttonRow: css({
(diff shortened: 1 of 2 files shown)


components/Trial/Form.js
```diff
@@ -33,6 +39,23 @@ const styles = {
     marginTop: 40,
     width: 170,
     textAlign: 'center'
+  }),
+  successActions: css({
+    display: 'flex',
+    flexDirection: 'column',
+    [mediaQueries.mUp]: {
+      display: 'inherit'
+    }
```
tpreusse: Alternative solution:
```suggestion
    },
    '& :first-child': {
      margin: '0 0 10px'
    },
    [mediaQueries.mUp]: {
      '& :first-child': {
        margin: '0 10px 0 0'
      }
    }
```
Then we also don't need prop forwarding on button.

annatraussnig: Duly noted!

PR #365, opened 2019-11-08
feat(Article): Propagate a payload w/ repoId, variation to trial form

patrickvenetz:
Depends on https://github.com/orbiting/backends/pull/331

Trial form accepts an optional `payload` which is submitted [upon requesting a trial](https://github.com/orbiting/republik-frontend/compare/feat-access-payload?expand=1#diff-41c2fe4a8031e9a159ffb7b63083d5d4R302-R306).

`payload` is a GraphQL input object, [defined in backends access package](https://github.com/orbiting/backends/pull/331/files#diff-55ab860d0141f0e282ee5f2d93b2d2b3) instead of a random JSON object we originally agreed on.

You may set a dedicated campaign for pay note trial flow with `TRIAL_PAYNOTE_CAMPAIGN` environment variable. Defaults to `TRIAL_CAMPAIGN`.


components/Trial/Form.js
```diff
@@ -98,6 +98,7 @@ const Form = props => {
   const [tokenType, setTokenType] = useState('EMAIL_CODE')
   const [showErrors, setShowErrors] = useState(false)
   const [autoRequestAccess, setAutoRequestAccess] = useState(false)
+  const [payload] = useState(props.payload)
```
tpreusse: This is actually not necessary and could lead to bugs.

E.g. when navigating from Episode 1 to Episode 2 the pay note could potentially not be remounted and the old payload would persist.

Directly using `props.payload` should be fine because by the time the variation changes to thank you the form is no longer visible and the `requestAccess` mutation has already run with the correct variation.

lib/constants.js
```diff
@@ -67,4 +67,5 @@ export const PROGRESS_EXPLAINER_PATH = ENV.PROGRESS_EXPLAINER_PATH
 export const ONBOARDING_PACKAGES = ['ABO', 'BENEFACTOR', 'MONTHLY_ABO']
 
 export const TRIAL_CAMPAIGN = ENV.TRIAL_CAMPAIGN
+export const TRIAL_PAYNOTE_CAMPAIGN = ENV.TRIAL_PAYNOTE_CAMPAIGN
```
tpreusse: wouldn't it be more consistent to use `TRIAL_CAMPAIGNS`?

components/Trial/Form.js
patrickvenetz: Thank you, got it.

lib/constants.js
patrickvenetz: Yes-ish. Adopting.

PR #369, opened 2019-11-26
Sections

garamond:
**Deployment procedure:**
1. Deploy `republik-publikator-api`
2. Reindex document index
3. Update `TEMPLATE` environment variable in `publikator-frontend`
4. Deploy `republik-api`, `publikator-frontend`, `republik-frontend`


components/Sections/SectionFeed.js
```diff
@@ -0,0 +1,80 @@
...
+
+import Box from '../Frame/Box'
+import { onDocumentFragment as bookmarkOnDocumentFragment } from '../Bookmarks/fragments'
+import { WithoutMembership, WithActiveMembership } from '../Auth/withMembership'
+
+import DocumentListContainer from '../Feed/DocumentListContainer'
+
+const SectionFeed = ({ t, formats }) => {
+  const getFeedDocuments = `
+  query getBookmarkedDocuments($cursor: String) {
```
tpreusse:
```suggestion
  query getSectionDocuments($cursor: String, $formats: [String!]) {
```

Proper name & pass the formats as a proper variable. Better for backend query analysis.

components/Sections/SectionFeed.js
```diff
@@ -0,0 +1,80 @@
...
+      documents(formats: ["${formats.join(
+        '","'
+      )}"], first: 30, after: $cursor) {
```
tpreusse:
```suggestion
      )}"], feed: true, first: 30, after: $cursor) {
```

Only show documents with the feed flag.

components/Sections/SectionFeed.js
garamond: Good catch, thanks!

PR #381, opened 2020-01-02
add edit icon to Article-ActionBar

patte:
If the user `isEditor` an edit icon is shown in the Article-ActionBar which opens the corresponding tree view in publikator.
Publikator target must be set with the new ENV `PUBLIKATOR_BASE_URL=https://publikator.republik.ch`

Screenshot from 2020-01-02 16-10-32


patte:
For the sake of simplicity I didn't add it to the Feed-ActionBar. But I could add it if desired.

components/ActionBar/index.js
```diff
@@ -197,6 +213,12 @@ class ActionBar extends Component {
           {displayConsumptionMinutes > 1 && (
             <ReadingTime minutes={displayConsumptionMinutes} />
           )}
+          {isEditor &&
+            editorIcons
+              .filter(Boolean)
+              .map((props, i) => (
+                <IconLink key={props.icon} fill={fill} {...props} />
```
tpreusse: fill comes through props spread:
```suggestion
                <IconLink key={props.icon} {...props} />
```

PR #396, opened 2020-02-13
Notifications

annatraussnig:
_Faaaar from ready, just adding screens of how things look as they come for you to see and comment on._

### Menu

Mobile (Still pondering – getting a bit cramped up there…)
Screenshot 2020-03-05 at 16 25 13
Screenshot 2020-03-05 at 16 25 20

Desktop
Screenshot 2020-03-05 at 16 25 31

### Notifications Page

_Atm looks like feed_

New/unread notifications are highlighted:
Screenshot 2020-03-04 at 11 04 19

If a notification arrives while user is on the page, a banner gets displayed:
Screenshot 2020-03-04 at 11 05 06

Everything read:
Screenshot 2020-03-04 at 11 05 43

Prototype for unsubscribe UI:
Screenshot 2020-03-05 at 18 19 54

Screenshot 2020-03-05 at 18 18 15

### Discussion notifications

Desktop version
Screenshot 2020-03-05 at 12 18 35

Mobile version
Screenshot 2020-03-05 at 16 23 57


components/Notifications/NewsletterSubscriptions.js
```diff
@@ -6,7 +6,7 @@ import withT from '../../lib/withT'
 import ErrorMessage from '../ErrorMessage'
 
 import FrameBox from '../Frame/Box'
-import { P } from './Elements'
+import { P } from '../Account/Elements'
```
tpreusse: I would leave that file in the account folder – newsletter stay separate from notifications for now.

tpreusse:
### Design Review

#### Unread Icon

> Screenshot 2020-03-24 at 23 27 56

With an unread notification I'd like to use the filled icon (still with the red dot on top and a white outline). Then it's consistent all the way: `outline = none`, `filled = some`.

#### Callout on Desktop

> Screenshot 2020-03-24 at 22 59 44

- icon duplication in callout seems unnecessary (even smaller bell icon)
- font size 12 is way too small, should be at least 14. But why not the regular 16?
- font size in callout does not scale with rest, ideally would listen to font size preference
- cursive titles should be medium instead

### Specific Use Cases


#### Format Page

> Screenshot 2020-03-24 at 23 59 56

In this case title «Benachrichtigungen» should be medium and the checkbox label should just be «Abonnieren»

#### Article / Newsletter

> Screenshot 2020-03-24 at 23 31 38

Title: Rubrik abonnieren
Checkbox Label: Format Title as is

#### Discussion with Format

> Screenshot 2020-03-25 at 00 38 56

Make format checkbox first and like article:  
Title: Rubrik abonnieren
Checkbox Label: Format Title as is

Then a second medium title: Debattenbeiträge abonnieren  
and then the radios as is.

(not sure if we should keep e.g. «Via App und E-Mail» or just always do an settings link at the end)

PR #408, opened 2020-03-26
Komplizen Formular

annatraussnig:
Edits to accomplice questionnaire:

- Text edits
- Add newsletter signup in thank you screen
- Add accomplice role to new accomplices
- Remove obsolete isEligible logic from newsletter component in account

### DIFF
--- a/pages/questionnaireCrowd.js
+++ b/pages/questionnaireCrowd.js
@@ -194,63 +194,62 @@ const ThankYouItem = compose(withT)(({ t, tKey }) => {
   )
 })
 
-const ThankYou = compose(withT)(({ t }) => {
+const ThankYou = compose(
+  withT,
+  graphql(NEWSLETTER_SETTINGS)
+)(({ t, data: { loading, error, me } }) => {
   return (
     <div>
       <Headline>{t('questionnaire/crowd/submitted/title')}</Headline>
       <div {...styles.intro}>
         <P>{t('questionnaire/crowd/submitted/intro')}</P>
       </div>
-      <Query query={NEWSLETTER_SETTINGS}>
-        {({ loading, error, data }) => {
-          if (loading || error) {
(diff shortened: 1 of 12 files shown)


pages/questionnaireCrowd.js
```diff
@@ -273,17 +356,20 @@ class QuestionnaireCrowdPage extends Component {
...
     const hasErrors = this.processErrors(errorMessages)
     if (hasErrors) return
 
     this.setState({ updating: true })
     this.processSubmit()
-      .then(() =>
-        this.setState(() => ({
+      .then(() => {
+        willingnessStatus === 'true' && addMeToRole({ role: 'accomplice' })
+        return this.setState(() => ({
```
tpreusse: We should add `addMeToRole` to the promise chain: ensuring the error is handled.
```suggestion
      .then(() => {
        if (willingnessStatus === 'true') {
          return addMeToRole({ role: 'accomplice' })
        }
      })
      .then(() => {
        return this.setState(() => ({
```

tpreusse: also after `addMeToRole` we should also auto subscribe to the NL in a additional `then`.

annatraussnig: Shoud we? I spoke with Carla on Thursday and we agreed to the show the subscription checkbox in the thank you screen and let the people decide if they want to subscribe or not.

tpreusse: My opinion ist yes: if they don't even want a newsletter about helping us initially, why even bother helping us? But practically it should be linked to the question «Wie dürfen wir Sie kontaktieren? Wählen Sie so viele Optionen, wie Sie wollen.». I didn't realize that they still have that one and thereby that they make it optional to be contacted by email. This is IMO a mistake on their part.

PR #415, opened 2020-04-21
added ability to scroll to progess from action bar

ovbm:
wrapped progress indicator in action bar in button to enable scroll to progress even if user closes the progress overlay.
ezgif-4-912dad66cfd4

### DIFF
--- a/components/ActionBar/Article.js
+++ b/components/ActionBar/Article.js
@@ -4,7 +4,7 @@ import ActionBar from './'
 import DiscussionIconLink from '../Discussion/IconLink'
 import { getDiscussionIconLinkProps } from './utils'
 
-import { fontStyles, button, plainButtonRule } from '@project-r/styleguide'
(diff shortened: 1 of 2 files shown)


components/ActionBar/Article.js
```diff
@@ -1,46 +1,50 @@
 import React, { Component, Fragment, useState, useEffect } from 'react'
+import PropTypes from 'prop-types'
 import ActionBar from './'
 import DiscussionIconLink from '../Discussion/IconLink'
 import { getDiscussionIconLinkProps } from './utils'
 
-import { fontStyles } from '@project-r/styleguide'
+import { fontStyles, button, plainButtonRule } from '@project-r/styleguide'
```
tpreusse, line 7:
```suggestion
import { fontStyles, plainButtonRule } from '@project-r/styleguide'
```
The lower case button is just the primitiv web button that does not need to be imported.

PR #428, opened 2020-06-10
Moved audioplayer out of header and into fixed component

ovbm:
localhost_3010_2020_03_26_am-klavier-kleinigkeiten(iPhone 5_SE)

localhost_3010_2020_03_26_am-klavier-kleinigkeiten(Laptop with MDPI screen)

-[ ] To test: iOS Safari tapping on seeker, does it open Safari Menu Bar?

### DIFF
--- a/components/Audio/index.js
+++ b/components/Audio/index.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect, useCallback } from 'react'
 import { AudioPlayer } from '@project-r/styleguide'
 import withT from '../../lib/withT'
 import { compose } from 'react-apollo'
@@ -16,7 +16,6 @@ export const AudioProvider = ({ children, t }) => {
   const [audioPlayerVisible, setAudioPlayerVisible] = useState(false)
 
   const toggleAudioPlayer = payload => {
-    setAudioPlayerVisible(true)
     setAudioState(payload)
   }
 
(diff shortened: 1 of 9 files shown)


components/Audio/index.js
```diff
@@ -0,0 +1,89 @@
...
+    setAudioPlayerVisible(true)
+    setAudioState(payload)
+  }
+
+  const onCloseAudioPlayer = () => {
+    setAudioPlayerVisible(false)
+    setTimeout(() => {
+      setAudioState(undefined)
+    }, 300)
+  }
```
tpreusse, line 28: Move the `{audioState && (` up around the first `div`. And then do the following:

```suggestion
  const [audioState, setAudioState] = useState(undefined)
  const [audioPlayerVisible, setAudioPlayerVisible] = useState(false)

  const toggleAudioPlayer = payload => {
    setAudioState(payload, () => {
      setAudioPlayerVisible(true)
    })
  }
  const onCloseAudioPlayer = () => {
    setAudioPlayerVisible(false)
  }
  
  useEffect(() => {
    if (!audioPlayerVisible && audioState) {
      const timeout = setTimeout(() => {
        setAudioState(undefined)
      }, 300)
      return () => {
        clearTimeout(timeout)
      }
    }
  }, [audioState, audioPlayerVisible])
```

The `useEffect` ensures that if the `audioState` changes in the 300ms fade out that `setAudioState(undefined)` is no longer called.

ovbm:
Hab nen weg gefunden mit useEffect, sollte eigentlich gehen so.

PR #431, opened 2020-06-19
Header & Navigation

ovbm:
TODO
- [x] IE 11 compatibility


patrickvenetz:
giphy

components/Article/Page.js
```diff
@@ -622,6 +626,18 @@ class ArticlePage extends Component {
...
+          animate: false,
+          estimatedReadingMinutes: undefined,
+          estimatedConsumptionMinutes: undefined,
+          onPdfClick: undefined,
+          pdfUrl: undefined,
+          showSubscribe: false,
+          fontSize: false,
+          wrapped: true
+        })
+      : undefined
```
tpreusse, line 640: Instead of having `actionBarNavNew` and `actionBarNav` const it would be better to just have one and configure it differently depending on the `isTester` flag:

```suggestion
    const actionBarNav = actionBar
      ? React.cloneElement(actionBar, isTester ? {
          animate: false,
          estimatedReadingMinutes: undefined,
          estimatedConsumptionMinutes: undefined,
          onPdfClick: undefined,
          pdfUrl: undefined,
          showSubscribe: false,
          fontSize: false,
          wrapped: true
        } : {
          animate: false,
          estimatedReadingMinutes: undefined,
          estimatedConsumptionMinutes: undefined,
          onPdfClick: undefined,
          pdfUrl: undefined,
          showSubscribe: false
        })
      : undefined
```

Currently you accidentally use `actionBarNavNew` even with the old nav.

components/Bookmarks/BookmarkMiniFeed.js
```diff
@@ -0,0 +1,116 @@
...
+                estimatedReadingMinutes,
+                description,
+                path
+              } = node.document.meta
+              return (
+                <div {...styles.tile} key={node.id}>
+                  <Link path={path} passHref>
+                    <div
+                      {...styles.tileHeadline}
+                      onClick={() => console.log('click on first tile')}
```
tpreusse, line 35: rm
```suggestion
                      onClick={() => console.log('click on first tile')}
```

components/Bookmarks/BookmarkMiniFeed.js
```diff
@@ -0,0 +1,116 @@
...
+              const { userProgress, userBookmark } = node.document
+              const {
+                estimatedReadingMinutes,
+                description,
+                path
+              } = node.document.meta
+              return (
+                <div {...styles.tile} key={node.id}>
+                  <Link path={path} passHref>
+                    <div
```
tpreusse, line 33: this should become an a tag (otherwise `passHref` make no sense and is not cmd clickable)

components/Bookmarks/BookmarkMiniFeed.js
```diff
@@ -0,0 +1,116 @@
...
+                      {...styles.tileHeadline}
+                      onClick={() => console.log('click on first tile')}
+                    >
+                      {description.substring(0, 50)}
+                      {description.length >= 50 ? '...' : ''}
```
tpreusse, line 38: Stylistically we do space + U2026 (horizontal ellipsis symbol). And non-breakable to prevent the dots of being on a line alone:
```suggestion
                      {description.substring(0, 50).trim()}
                      {description.length >= 50 && <>&nbsp;…</>}
```

components/Feed/index.js
```diff
@@ -62,88 +62,79 @@ const greetingSubscription = gql`
...
-  }
+  }, [subscribeToMore])
 
-  render() {
-    const {
-      meta,
-      data: { error, loading, greeting, documents: connection, fetchMore }
-    } = this.props
+  return (
+    <Frame hasOverviewNav={true} raw meta={meta}>
```
tpreusse, line 107: Generally we use the short hand for bool flags:
```suggestion
    <Frame hasOverviewNav raw meta={meta}>
```

components/Frame/HLine.js
```diff
@@ -0,0 +1,64 @@
...
+import {
+  HEADER_HEIGHT,
+  HEADER_HEIGHT_MOBILE,
+  ZINDEX_POPOVER
+} from '../constants'
+
+const HLine = ({ formatColor, dark }) => {
+  const hrColor = dark ? colors.negative.divider : colors.divider
+  const hrColorStyle = { color: hrColor, backgroundColor: hrColor }
+  console.log(formatColor)
```
tpreusse, line 14: rm

components/Frame/SecondaryNav.js
```diff
@@ -0,0 +1,161 @@
...
+  isFront,
+  showSecondary,
+  router,
+  hasOverviewNav,
+  t
+}) => {
+  const active = matchPath(router.asPath)
+  return (
+    <>
+      {isFront || hasOverviewNav ? (
```
tpreusse, line 32: You should actually set `hasOverviewNav` on the front page:

```suggestion
      {hasOverviewNav ? (
```

components/Frame/SecondaryNav.js
```diff
@@ -0,0 +1,161 @@
...
+              hoverColor={colors.primary}
+              minifeed={true}
+            >
+              {t('navbar/discussion')}
+            </NavLink>
+            <Sections dark={dark} active={active} minifeed={true} />
+          </div>
+        </div>
+      ) : (
+        // !!showSecondary &&
```
tpreusse, line 70: needs clean up

components/Frame/indexNew.js
```diff
@@ -0,0 +1,189 @@
+import React from 'react'
```
tpreusse, line 1: this this actually used?

components/Front/index.js
```diff
@@ -166,7 +166,7 @@ const Front = ({
   }
 
   return (
-    <Frame raw meta={meta}>
+    <Frame hasOverviewNav={true} raw meta={meta}>
```
tpreusse, line 169:
```suggestion
    <Frame hasOverviewNav raw meta={meta}>
```

components/Gallery/Gallery.js
```diff
@@ -85,7 +85,11 @@ const Gallery = ({ items, onClose, startItemSrc, children, t }) => {
         className='pswp'
         tabIndex='-1'
         role='dialog'
-        style={{ background: '#000' }}
+        style={{
+          background: '#000',
+          top: isTester && headerHeight,
```
tpreusse, line 90: Why did you change this behaviour? I think the gallery should keep going over the header – we should keep the fullscreen image experience free of headers.

components/Icons/SearchMenu.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+
+const Icon = ({ size, fill, ...props }) => (
+  <svg width={size} height={size} viewBox='0 0 26 18' {...props}>
+    <path
+      fillRule='evenodd'
+      clipRule='evenodd'
```
tpreusse, line 7: clip rule is unnecessary:
```suggestion
```

I generally would also recommend running icons through [SVGOMG](https://jakearchibald.github.io/svgomg/) (would reduce the number of paths elements in this cases).

components/Gallery/Gallery.js
ovbm: reverted

components/Frame/indexNew.js
ovbm: nope, deleted

components/Icons/SearchMenu.js
ovbm: done

components/Frame/Popover/UserNav.js
```diff
@@ -0,0 +1,279 @@
...
+  smallLinks: css({
+    '& a': {
+      ...fontStyles.sansSerifRegular18
+    }
+  })
+}
+
+export default compose(
+  withT,
+  graphql(notificationsQuery),
```
tpreusse, line 276: There is a loading explosion on love because loading and error state of this query aren't handled.

Move this to `NotificationFeedMini` and properly handle the loading state there (with a `<Loader />` like book mark mini feed, btw it's once `MiniFeed` and once `FeedMini`, I'm in favor of mini feed)

tpreusse:
Current IE11 looks:
Screenshot 2020-06-26 at 15 10 43

Let's review the new css for the header and mini menu together next week.

components/Frame/Popover/UserNav.js
ovbm: done 🧞

components/Bookmarks/BookmarkMiniFeed.js
```diff
@@ -0,0 +1,121 @@
...
+const BookmarkMiniFeed = ({ t, data, ...props }) => {
+  return (
+    <Loader
+      loading={data.loading}
+      error={data.error}
+      render={() => {
+        const nodes = data.me.collection.items.nodes
+        return (
+          <div {...styles.tilesContainer} {...props}>
+            {nodes.slice(0, 8).map(node => {
```
tpreusse, line 29: `node.document` can be null (e.g. staging, doc not available or unpublished). Let's just filter those out:
```suggestion
            {nodes.filter(node => node.document).slice(0, 8).map(node => {
```

components/Frame/HLine.js
ovbm: done

components/Frame/SecondaryNav.js
```diff
@@ -0,0 +1,158 @@
...
+      justifyContent: 'center'
+    },
+    '& a': {
+      display: 'block',
+      whiteSpace: 'nowrap',
+      fontSize: 14,
+      margin: '0 16px',
+      ':active': {
+        fontWeight: 'bold'
+      },
```
tpreusse, line 126: You should never manually set `fontWeight` or just `fontFamily` but always use `fontStyles` which contains a valid combinations of those two. In this case you can directly assign:
```suggestion
      ':active': fontStyles.sansSerifMedium,
```

(If you had additional styles you could spread it: `{ ...fontStyles.sansSerifMedium, color: 'red' }`)

ovbm: jo hab ich sonst immer gemacht, weiss nicht warum hier nicht XD

components/Frame/HeaderNew.js
```diff
@@ -0,0 +1,387 @@
...
+          (showSecondary ? SUBHEADER_HEIGHT : 0) +
+          headerOffset
+      },
+      {
+        minWidth: mediaQueries.mBreakPoint,
+        headerHeight:
+          HEADER_HEIGHT + (showSecondary ? SUBHEADER_HEIGHT : 0) + headerOffset
+      }
+    ]
+  }, [secondaryNav, headerOffset])
```
tpreusse, line 177: The header height context is currently wrong when `hasOverviewNav`.

```suggestion
  const headerConfig = useMemo(() => {
    return [
      {
        minWidth: 0,
        headerHeight:
          HEADER_HEIGHT_MOBILE +
          (hasOverviewNav || secondaryNav ? SUBHEADER_HEIGHT : 0) +
          headerOffset
      },
      {
        minWidth: mediaQueries.mBreakPoint,
        headerHeight:
          HEADER_HEIGHT + (hasOverviewNav || secondaryNav ? SUBHEADER_HEIGHT : 0) + headerOffset
      }
    ]
  }, [hasOverviewNav, secondaryNav, headerOffset])
```

Or refactor that `hasOverviewNav` actually sets `secondaryNav`.

components/Frame/Popover/NavNew.js
```diff
@@ -0,0 +1,225 @@
...
+                <Link
+                  href='/sections'
+                  active={active}
+                  closeHandler={closeHandler}
+                  passHref
+                >
+                  <TeaserSectionTitle small>
+                    {t('nav/sections')}
+                  </TeaserSectionTitle>
+                </Link>
```
tpreusse, line 108: The sections/Rubriken link is currently broken.

You're using the wrong link here. Remove `import Link from '../../Link/Href'` and import `Link` from `lib/routes`: `import { matchPath, Link } from '../../../lib/routes'`.

And try this:

```suggestion
                <Link
                  route='sections'
                  passHref
                >
                  <TeaserSectionTitle small onClick={closeHandler}>
                    {t('nav/sections')}
                  </TeaserSectionTitle>
                </Link>
```
(not a 100% sure that the onClick is triggered but if I remember our last test together correctly it was)

The path link you were using never supported active or close handler props. Maybe you intended to use the `NavLink`? But it probably can't handle the style correctly.

components/Frame/Popover/UserNav.js
```diff
@@ -0,0 +1,254 @@
...
+                  href='https://www.republik.ch/pledge'
+                  black
+                  block
+                >
+                  {t('nav/becomemember')}
+                </Button>
+              )}
+              {me && (
+                <>
+                  <Link href='/benachrichtigungen' active={active} passHref>
```
tpreusse, line 89: Please also use the `Link` from `lib/routes` here. And remove the `active` prop – not supported on either of those – that a `NavLink` specific thing.

```suggestion
                  <Link route='subscriptions' passHref>
```

components/Frame/Popover/UserNav.js
```diff
@@ -0,0 +1,254 @@
...
+                    <TeaserSectionTitle onClick={() => closeHandler()} small>
+                      {t('pages/notifications/title')}
+                    </TeaserSectionTitle>
+                  </Link>
+                  <NotificationFeedMini closeHandler={closeHandler} />
+                  <br />
+                  <Link href='/lesezeichen' active={active} passHref>
```
tpreusse, line 96: see above

components/Notifications/NotificationFeedMini.js
```diff
@@ -0,0 +1,143 @@
...
+                    }}
+                  />
+                </Interaction.P>
+              </>
+            )}
+
+            {!newNodes.length && (
+              <p>
+                {t('pages/notifications/nounread')}{' '}
+                <Link route='/benachrichtigungen' passHref>
```
tpreusse, line 66:
```suggestion
                <Link route='subscriptions' passHref>
```

components/Frame/UserNew.js
```diff
@@ -0,0 +1,121 @@
...
+  )
+    .slice(0, 2)
+    .filter(Boolean)
+    .map(s => s[0])
+    .join('')
+
+const User = ({ t, me, title, dark, backButton, ...props }) => {
+  const color = dark ? colors.negative.text : colors.text
+  return (
+    <div {...styles.user} {...props}>
```
tpreusse, line 33: By spreading props here you assign the `onClick` handler to this div which does not have `role='button'` – role button and onClick should be on the same element. Generally it's better to assign explicits props that you support instead of forwarding rest props (creates uncertainty how this component is used – what other props are set on this div and why?).
```suggestion
    <div {...styles.user} {...props}>
```

components/Frame/HeaderNew.js
```diff
@@ -0,0 +1,387 @@
...
+  const lastY = useRef()
+  const headerHeight = useRef()
+  const lastDiff = useRef()
+
+  const textFill = dark ? colors.negative.text : colors.text
+  const logoFill = dark ? colors.logoDark || '#fff' : colors.logo || '#000'
+  const backButton = inNativeIOSApp && me && !isFront(router)
+
+  const toggleExpanded = target => {
+    if (target.id === expandedNav) {
```
tpreusse, line 87: why go trough the DOM? Instead of passing `e.currentTarget` you could just directly pass the id / nav name?

Currently clicking on the profile pic when the user nav is already open does not hide it. Some logic here is broken.

components/Frame/HeaderNew.js
```diff
@@ -0,0 +1,387 @@
...
+
+  useEffect(() => {
+    window.addEventListener('scroll', onScroll)
+    window.addEventListener('resize', measure)
+    measure()
+    return () => {
+      window.removeEventListener('scroll', onScroll)
+      window.removeEventListener('resize', measure)
+    }
+  }, [showSecondary])
```
tpreusse, line 160: this effect does not correctly scope dependencies and the on scroll and reize functions might use outdated state

components/Article/Page.js
```diff
@@ -6,7 +6,7 @@ import Frame from '../Frame'
 import ArticleActionBar from '../ActionBar/Article'
 import Loader from '../Loader'
 import RelatedEpisodes from './RelatedEpisodes'
-import SeriesNavButton from './SeriesNavButton'
+import SeriesNavButton from './SeriesNavButtonNew'
```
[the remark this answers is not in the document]
ovbm: swapped it in just to be safe.

components/Frame/Popover/NavNew.js
ovbm: TeaserSectionTitle accepts an onClick handle, which is how I used it in UserNav as well. so that should work

components/Frame/HeaderNew.js
ovbm: Fixed the DOM part. The case of the nav not closing on clicking the user icon when the user nav is already open is as intended. Both navs close via the close button on the right. I thought it could be perceived as inconsistent if suddenly the user icon would double as a close button, even though there is an explicit close button on the right.

components/Frame/UserNew.js
ovbm: done

components/Frame/HeaderNew.js
ovbm: moved onscroll and measure inside useeffect and added all dependencies

components/Gallery/Gallery.js
```diff
@@ -1,21 +1,20 @@
...
 import PhotoSwipe from 'photoswipe'
 import PhotoSwipeUIDefault from 'photoswipe/dist/photoswipe-ui-default'
-
 import { imageSizeInfo, imageResizeUrl } from 'mdast-react-render/lib/utils'
+import { Spinner } from '@project-r/styleguide'
+
+import withT from '../../lib/withT'
+import { withTester } from '../Auth/checkRoles'
+import photoswipeStyle from './photoswipeStyle'
+import { ZINDEX_LOADINGBAR } from '../constants'
```
tpreusse, line 11: why `ZINDEX_LOADINGBAR`? IMO dedicated const or `zIndex.overlay` from the [styleguide](https://github.com/orbiting/styleguide/blob/44aceafe67794d7bd112ecce6fe7547fabcbaca8/src/theme/zIndex.js#L5).

components/Gallery/Gallery.js
```diff
@@ -1,21 +1,20 @@
...
 
 const removeQuery = (url = '') => url.split('?')[0]
 
 const MAX_SPREAD_ZOOM = 2
 
-const Gallery = ({ items, onClose, startItemSrc, children, t }) => {
+const Gallery = ({ items, onClose, startItemSrc, children, t, isTester }) => {
```
tpreusse, line 17: unused `isTester` and `withTester`

PR #433, opened 2020-06-25
Enables User-Subscriptions

ovbm:


tpreusse:
One small note: please add translation to «[Republik Translations](https://docs.google.com/spreadsheets/d/1a-8RESf0Yjo2Mu15Q4SGWTmnY3ifuFbCVSD7WyYRhT0/edit#gid=0)» in Google drive.

After adding it there you can run `npm run translations` to add them to the json.

ovbm:
Ready for review
Tested Flexbox layout on IE11
Regarding translations: I'll add them retroactively to the Drive

components/Profile/Page.js
```diff
@@ -565,6 +582,13 @@ const LoadedProfile = props => {
                   errors={errors}
                   dirty={dirty}
                 />
+                <div>
+                  <SubscribeMenu
+                    label='Profil Abonnieren'
```
tpreusse, line 587: add to translations and AFAIK should be `Profil abonnieren`

components/Profile/Page.js
```diff
@@ -565,6 +582,13 @@ const LoadedProfile = props => {
...
+                    showAuthorFilter={true}
+                    subscriptions={{ nodes: Array.of(user.subscribedByMe) }}
```
tpreusse, line 589:
```suggestion
                    subscriptions={{ nodes: [user.subscribedByMe] }}
```
(I did end up checking and we do have the IE11 polyfill via next.js but I prefer using syntax instead of methods when possible)

components/Notifications/SubscribedAuthors.js
```diff
@@ -0,0 +1,139 @@
...
+    <Loader
+      loading={loading}
+      error={error}
+      render={() => {
+        return (
+          <>
+            {!isMember && (
+              <Box style={{ margin: '10px 0', padding: 15 }}>
+                <Interaction.P>
+                  {t('Notifications/settings/formats/noMembership')}
```
tpreusse, line 83: Maybe we could reformulate this message to work for formats and authors and move it above both components? (in `components/Notifications/Settings.js`)

components/Article/Page.js
```diff
@@ -143,6 +143,11 @@ const getDocument = gql`
       subscribedByMe(includeParents: true) {
         ...subInfo
       }
```
tpreusse, line 145: could this already be removed?

components/Article/Page.js
```diff
@@ -614,7 +619,9 @@ class ArticlePage extends Component {
         showBookmark={isMember}
         estimatedReadingMinutes={meta.estimatedReadingMinutes}
         estimatedConsumptionMinutes={meta.estimatedConsumptionMinutes}
+        // TODO: replace this code
         subscription={article.subscribedByMe}
```
tpreusse, line 623: could this already be removed?

components/Article/Page.js
```diff
@@ -643,6 +650,18 @@ class ArticlePage extends Component {
               }
         )
       : undefined
+    const actionBarNavNew = actionBar
```
tpreusse, line 653: rm?

components/Article/Page.js
ovbm: why should this be removed? It's used further down

components/Notifications/SubscribeCheckbox.js
```diff
@@ -0,0 +1,119 @@
...
+    : subscription && subscription.active
+  const activeFilters = (subscription.active && subscription.filters) || []
+  const isDocument =
+    subscription &&
+    subscription.object &&
+    subscription.object.__typename === 'Document'
+
+  const toggleCallback = () => setAnimate && setAnimate(true)
+
+  const toggleSubscribe = () => {
```
tpreusse, line 57: we should handle loading and error of the mutation: add a new `isLoading`/`isMutating` state which sets ` <Checkbox disabled />` and a `serverError` state which sets `<Checkbox error />` and potential prints the message somewhere, but just marking the error is a good first step. You can get the error with a `.catch` after the `.then`.

PR #434, opened 2020-06-30
Submit Pledge Payload

tpreusse:
Integrate new pledge payload https://github.com/orbiting/backends/pull/449

Send campaign referrers as a first step.

### DIFF
--- a/lib/utils/url.js
+++ b/lib/utils/url.js
@@ -1,4 +1,19 @@
-export const getUtmParams = (query, defaults = {}) =>
+export const getUtmParams = query => {
+  // support /probelesen?campaign=x
+  let defaultCampaign = query.campaign
+  // check matomo attribution cookie
+  if (typeof window !== 'undefined' && window.Matomo) {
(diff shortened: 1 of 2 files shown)


lib/utils/url.js
```diff
@@ -1,4 +1,19 @@
-export const getUtmParams = (query, defaults = {}) =>
+export const getUtmParams = query => {
+  // support /probelesen?campaign=x
+  let defaultCampaign = query.campaign
+  // check matomo attribution cookie
+  if (typeof window !== 'undefined' && window.Matomo) {
```
patrickvenetz, line 5: Since you're accessing a property afterwards, aim for a more specific genesis test like `typeof window !== 'object'`?

tpreusse: `typeof window === 'object'` would be an alternative but not only objects can have properties (e.g. function can also have them). I prefer the `undefined` check because I do not assume that browser implement it as object. On the other hand they seem to always do that.

PR #435, opened 2020-07-03
Navigation Tweaks (feedback)

ovbm:
Before:
Screenshot 2020-07-03 at 10 09 20

After:
Screenshot 2020-07-03 at 10 09 02


ovbm:
before:
Screenshot 2020-07-03 at 10 54 07

after:
Screenshot 2020-07-03 at 10 53 53

PR #443, opened 2020-07-10
Main Nav Accordion

ovbm:


components/Frame/Popover/NavLink.js
```diff
@@ -45,6 +47,20 @@ export const NavA = React.forwardRef(
...
+          },
+          ...fontStyles.sansSerifMedium20,
+          [mediaQueries.mUp]: {
+            ...fontStyles.sansSerifMedium22,
+            '& + &': {
+              marginLeft: 36,
+              marginTop: 0
+            }
+          }
+        }))}
```
tpreusse, line 63: Move this to `styles` as a `large` key and then do `styles.large` here. Currently on every render of every instance it has to compute a css hash and then realise that it already generated the css rule.

package.json
```diff
@@ -103,7 +103,7 @@
   },
   "dependencies": {
     "@next/bundle-analyzer": "^9.4.4",
-    "@project-r/styleguide": "^9.19.6",
+    "@project-r/styleguide": "https://s3.eu-central-1.amazonaws.com/republik-assets-dev/npm/project-r-styleguide-ob-0c08b60.tgz",
```
tpreusse, line 106: bump styleguide to `^9.20.0` :)

PR #456, opened 2020-07-29
3rd Wave

tpreusse:
Depends on
- https://github.com/orbiting/backends/pull/478
- https://github.com/orbiting/backends/pull/481

Change log
- enable and disable auto pay in account 
- show credit card form when auto pay is on
- refactor membership actions
  - limit all membership actions to active memberships except prolong
- disable cancel form when `membership.renew` is `false`

Screenshot 2020-07-29 at 14 41 03

Screenshot 2020-07-29 at 14 56 03

### DIFF
--- a/components/Account/index.js
+++ b/components/Account/index.js
@@ -82,7 +82,7 @@ class Account extends Component {
       hasActiveMemberships,
       hasAccessGrants,
       acceptedStatue,
-      recurringAmount,
+      isAutoPaying,
       hasPledges,
       merci,
       inNativeIOSApp,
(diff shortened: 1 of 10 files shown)


components/Account/index.js
```diff
@@ -226,16 +224,16 @@ export default compose(
...
-          : 0
+        isAutoPaying:
+          hasMemberships &&
+          data.me.memberships.some(m => {
+            if (!m.active || !m.renew) {
+              return false
+            }
+            if (m.type.name === 'MONTHLY_ABO' || m.autoPay) {
+              return true
+            }
```
patrickvenetz, line 235: Assuming it has to be active and renewable, either as a MONTHLY_ABO or with auto-pay-flag set to true, then condition could be:

```suggestion
            return m.active && m.renew && (m.type.name === 'MONTHLY_ABO' || m.autoPay)
```

(Changes untested.)

PR #460, opened 2020-08-18
New ActionBar

ovbm:
New Action Bar:
- Consistent icons and components (Use of styleguide IconButton instead of mix of IconLink and custom components)
- Mark as unread functionality
- ActionBar overlay on scroll-up

Article:
Screenshot 2020-08-20 at 14 34 50
Screenshot 2020-08-20 at 14 38 01

Scroll-up Overlay:
Screenshot 2020-08-20 at 14 37 19
Screenshot 2020-08-20 at 14 38 24

Feed:
- Removed functionality: gallery icon isn't shown in feed anymore
Screenshot 2020-08-20 at 14 39 38
Screenshot 2020-08-20 at 14 39 27

Known Issues
- [x] UserProgress: restoreArticleProgress scroll is jumpy and doesn't scroll to right position in ActionBarOverlay


components/ActionBar/index.js
```diff
@@ -1,281 +1,422 @@
...
+  const emailSubject = t('article/share/emailSubject', {
+    title: document.title
+  })
+  const {
+    discussionId,
+    discussionPath,
+    discussionQuery,
+    discussionCount,
+    isDiscussionPage
+  } = getDiscussionIconLinkProps(
```
ovbm, line 115: Any idea why this function always returns "/dialog" instead of the link to the discussion? I didn't change the function, just copied it, like it was done in the previous component. @tpreusse

components/Gallery/Gallery.js
```diff
@@ -70,14 +70,8 @@ const Gallery = ({ items, onClose, startItemSrc, children, t }) => {
...
-      gallery.listen('close', () => {
-        // workaround bug opening nav when closing gallery on touch devices
-        // - defer onClose one render frame (1000ms/30frames)
-        // - othwise the component would already be unmounted when the click reaches the React event system and trigger a event on whatever is underneath
-        setTimeout(() => {
-          onClose()
-        }, 33)
-      })
+      gallery.listen('close', e => console.log('close', e))
+      gallery.listen('pswpTap', e => console.log('tap', e, e.detail))
```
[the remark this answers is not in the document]
ovbm: I have no idea why this was changed XD

components/Icons/MdCheckCircleOutlined.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+export default ({ size = 24, fill }) => (
```
tpreusse, line 3: Next / the hot reloader does not like those anymore:
> Anonymous arrow functions cause Fast Refresh to not preserve local component state.
> Please add a name to your function, for example:
>
> Before
> export default () => <div />;
> 
> After
> const Named = () => <div />;
> export default Named;

Same with the other two icons.

components/Article/Page.js
```diff
@@ -471,50 +469,22 @@ class ArticlePage extends Component {
...
+
+    const actionBar = article && (
+      <ActionBar mode='article-top' document={article} />
     )
     const actionBarEnd = actionBar
       ? React.cloneElement(actionBar, {
-          animate: false,
-          estimatedReadingMinutes: undefined,
-          estimatedConsumptionMinutes: undefined,
-          grandSharing: !inNativeApp
```
tpreusse, line 517: Would it be hard to bring back `grandSharing` at the end of an article? This currently causes the following share buttons at the end:
Screenshot 2020-08-27 at 16 28 35
 
(not shown in the app for consistency because share triggers the native share dialog there)

We seem to already have a similar display with the new `IconButton` on the marketing page:
Screenshot 2020-08-27 at 16 27 59

Could we use that underneath the article too?

tpreusse (approved):
Great job on cleaning this up and the new on scroll action bar.

General approval from my side. A few notes here and there.

tpreusse:
One small idea: could we show email and copy link first (before the social networks) in share buttons? There are used the most.

components/Article/Page.js
ovbm: jup, shouldn't be a big change.

ovbm:
Screenshot 2020-08-27 at 17 16 33
done

tpreusse:
> done, appears when `mode === 'article-bottom'`, is there more fine grained control needed?

Well in theory I'd not show it in the app (`inNativeApp`) because the regular share button triggers the native dialog there. But maybe we should try and wait for feedback. Probably those direct links will work nicely too. Only real issue I see is that some people have a custom thingy in the native share panel that they want.

PR #477, opened 2020-09-18
Refactor of Article/Page.js

ovbm:
tested locally, next step, test on love


tpreusse (approved):
LGTM. Tested it locally with prd data and couldn't find anything.

PR #480, opened 2020-09-30
Filter Bookmark Page

ovbm:
Current State:
ezgif-6-8dc320252efa

Issues:
- [x] deduplication is not working
- [x] filter is not working properly. On first load it's as intended. But when clicking on "Gelesen" and going back to "Weiterlesen" the selection is not correct.
- [x] pagination?

### DIFF
--- a/components/Bookmarks/Page.js
+++ b/components/Bookmarks/Page.js
@@ -5,6 +5,7 @@ import Frame from '../Frame'
 import { enforceMembership } from '../Auth/withMembership'
 import DocumentListContainer from '../Feed/DocumentListContainer'
 import withT from '../../lib/withT'
+import withMe from '../../lib/apollo/withMe'
 
 import {
   mediaQueries,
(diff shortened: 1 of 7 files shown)


components/Bookmarks/Page.js
```diff
@@ -11,13 +11,164 @@ import {
...
+          <button
+            onClick={() => handleFilterClick('continue')}
+            {...plainButtonRule}
+            {...styles.filterItem}
+          >
+            <Interaction.H3
+              title='Weiterlesen'
+              {...styles.fiterItemText}
+              style={{
+                fontWeight: filter === 'continue' ? 'bold' : 'normal'
```
tpreusse, line 102: We need to always use `fontStyles` instead of setting `fontWeight` directly (because familyName, fontWeight and fontStyle all need to align to work properly). Also currently we don't habe sans serif bold, only medium. But can we align it with debates and search where we use underline? (see «Video» and «Beliebteste»)
Screenshot 2020-09-30 at 15 57 56

Screenshot 2020-09-30 at 15 57 35

ovbm: done

ovbm: Screenshot 2020-09-30 at 16 03 53

ovbm:
ready for review @tpreusse

tpreusse (approved):

PR #482, opened 2020-10-05
Meine Republik Tweaks

ovbm:
- Keine Callouts in "Meine Republik" Actionbar
- Filter: "Gemerkt" statt "Lesezeichen"
- MiniFeed: Weiterlesen statt Lesezeichen

### DIFF
--- a/components/Bookmarks/BookmarkMiniFeed.js
+++ b/components/Bookmarks/BookmarkMiniFeed.js
@@ -10,7 +10,8 @@ import Link from '../Link/Path'
 import { getCollectionItems, registerQueryVariables } from './queries'
 
 const variables = {
-  collections: ['bookmarks']
+  collections: ['progress', 'bookmarks'],
+  progress: 'UNFINISHED'
 }
 
 const BookmarkMiniFeed = ({ data, closeHandler, style }) => {
--- a/components/Bookmarks/BookmarkMiniFeed.js
+++ b/components/Bookmarks/BookmarkMiniFeed.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 import { compose, graphql } from 'react-apollo'
 import { css } from 'glamor'
 import { colors, fontStyles, mediaQueries, Loader } from '@project-r/styleguide'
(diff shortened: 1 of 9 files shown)


components/Bookmarks/queries.js
```diff
@@ -30,6 +30,7 @@ export const getCollectionItems = gql`
         after: $cursor
         progress: $progress
         uniqueDocuments: true
+        lastDays: 30
```
tpreusse, line 33: should be a null-able variables and only set to 30 when in «weiterlesen» view

components/Bookmarks/BookmarkMiniFeed.js
```diff
@@ -10,7 +10,8 @@ import Link from '../Link/Path'
 import { getCollectionItems, registerQueryVariables } from './queries'
 
 const variables = {
-  collections: ['bookmarks']
+  collections: ['progress', 'bookmarks'],
+  progress: 'UNFINISHED'
```
tpreusse, line 14: should be moved into a wrapper around the whole thing which does a different query based on `progressConsent`:
```
  const hasProgress = !!(me?.progressConsent)
  const variables = useMemo(() => {
    if (hasProgress) {
        return {
          collections: ['progress', 'bookmarks'],
          progress: 'UNFINISHED',
          lastDays: 30
        }
    }
    return {
      collections: ['bookmarks']
    }
  }, [hasProgress])
  registerQueryVariables(variables)
  return <BookmarkMiniFeed variables={variables} />
```

And then use a [options function](https://www.apollographql.com/docs/react/api/react/hoc/#configoptions) in the `graphql` wrapper:
```
  graphql(getCollectionItems, {
    options: props => ({
      fetchPolicy: 'cache-and-network',
      variables: props.variables
    })
  })
```

components/Bookmarks/queries.js
ovbm: ah ja logo XD

tpreusse (approved):

components/Front/index.js
```diff
@@ -121,18 +120,15 @@ const Front = ({
...
-    [isTester]
+  const schema = useMemo(() =>
+    createFrontSchema({
+      Link: HrefLink,
+      CommentLink,
+      DiscussionLink,
+      ...withData,
+      ActionBar,
+      t
+    })
```
tpreusse, line 131: Watch out: we need an empty array here:
> If no array is provided, a new value will be computed on every render.
> [useMemo docs](https://reactjs.org/docs/hooks-reference.html#usememo)

Issue #465, opened 2020-08-28
Follow author UX improvement


annatraussnig:
## Bug Report

When following an author on https://www.republik.ch/benachrichtigungen/einstellungen, said author jumps to the top of the section, which leads to quite awkward interactions

**Steps to Reproduce**

Go to https://www.republik.ch/benachrichtigungen/einstellungen
<img width="916" alt="Screenshot 2020-08-28 at 15 31 01" src="https://user-images.githubusercontent.com/3907984/91566167-887a0480-e943-11ea-9550-71a330532cd9.png">
Pick an author
Check "Magazin"

**Actual Results**

Bam! the author jumps up (cumbersome if you also planned to click "Debatte")

**Possible Solutions**

Only move the checked authors to the top of the section after page reload.

dgnaegi:
Hi @annatraussnig 

I am currently in quarantine and as a pastime I'm participating at [hacktoberfest](https://hacktoberfest.digitalocean.com/) and practising my react skills. I opened a PR with a possible solution to this issue:

https://github.com/orbiting/republik-frontend/pull/487

I would appreciate a short review. If it's a fitting solution please add the "hacktoberfest-accepted" label to the PR :)

Best wishes
Daniel

annatraussnig:
Oh, sorry to hear about the quarantine @dgnaegi – I hope it's not driving you up the walls. Really great that you are helping out with the bug, though! Many thanks! I'll have a look later today then :)

annatraussnig:
Might be tomorrow or Thursday :/
Running a little bit tight today but let's see...

annatraussnig:
Issue solved :) Thank you so much for your work!

PR #487, opened 2020-10-09
Follow author UX improvement #465 

dgnaegi:
idea / concept:
Initialize subscribed authors at first page rendering / "load more"-button-click and use the initialized data as sorting criteria

behavior


components/Notifications/SubscribedAuthors.js
```diff
@@ -153,3 +176,20 @@ const SubscribedAuthors = ({
 }
 
 export default compose(withT, graphql(myUserSubscriptions))(SubscribedAuthors)
+
+function InitializeSubscribedAuthorIds(
```
annatraussnig, line 180: The function name I'd start with lowercase. To be more consistent with the rest of the codebase, I'd also bring it inside the component and use `setInitialySubscribedAuthorIds`directly. That way, the `export default`also remains the last line of the file, which is also consistent with our codebase.

components/Notifications/SubscribedAuthors.js
```diff
@@ -65,6 +65,19 @@ const SubscribedAuthors = ({
...
+    setInitiallySubscribedAuthorIds
+  ] = useState([])
+
+  useEffect(() => {
+    InitializeSubscribedAuthorIds(
+      authors,
+      myUserSubscriptions.subscribedTo.nodes,
+      setInitiallySubscribedAuthorIds
+    )
+  }, [])
```
annatraussnig, line 79: Don't we run into the risk that `authors` and `myUserSubscriptions`are still undefined at this point?

annatraussnig (commented):
Hey, first thank you so much again! I just tried to run the code locally and faced 2 issues:

1. undefined authors error upon first loading the user preferences
2. the sorting in the expanded list is upside down (followed authors at the bottom -> I made a screenshot)

Screenshot 2020-10-19 at 16 26 13

If you have to time to have a look, that would be very cool indeed. If not, let me know so I'm aware and I can take over.

dgnaegi:
> If you have to time to have a look, that would be very cool indeed. If not, let me know so I'm aware and I can take over.

Thanks for the review :) Regarding the issues:

1) I didnt noticed this one :-/ Under the assumption that the list is intitially collapsed I could remove the useeffect completely so the list gets initiallized after expanding. But this could make further adjustments more complex. Would this be a solution? I already pushed the change but didnt test it yet. Will do so in the next days

2) On my machine, the sorting is correctly:

grafik

Could it be an enviroment issue? I noticed that on your screenshot there are some magazine-checkboxes missing as well compared to the production environment. Why is that?

On a logical side it is: Included in the list -> true -> 1 -> In descending order on top of 0

But i could dig into this issue as well, probably on thursday

annatraussnig:
Re 2. yes, those profile are still from Wahltindär. Are you connecting to the live DB or what's your setup? If you want to try and follow some the politicians and see if this plays a role, here are a couple of randomly chosen links:
https://www.republik.ch/~rsauter
https://www.republik.ch/~joergmaeder
https://www.republik.ch/~bwalti
I'm curious if this changes the game!

dgnaegi:
> Re 2. yes, those profile are still from Wahltindär. Are you connecting to the live DB or what's your setup? If you want to try and follow some the politicians and see if this plays a role, here are a couple of randomly chosen links:
> https://www.republik.ch/~rsauter
> https://www.republik.ch/~joergmaeder
> https://www.republik.ch/~bwalti
> I'm curious if this changes the game!

Please excuse the late reply, I was pretty busy the last couple of days. Thanks for the hint with the wahltindär, but this doesn't change the behavior on my machine. I do work with the Production-API's

grafik

You don't have to invest to much time into my PR, but im still wondering why we experience a different behavior...

annatraussnig (approved):

annatraussnig:
Hey Daniel! Thanks for your patience 😸 I also got busy with Urabstimmung work... Anyway thanks for the fixes, all seems to work now! I can merge the PR on Friday morning. It's a bit late for hacktoberfest… although you did most of the work in October… Shall I add the tag nonetheless or doesn't matter anymore?

PR #493, opened 2020-10-30
Dialog & Search Darkmode

ovbm:
Requires https://github.com/orbiting/styleguide/pull/335

Dialog
Screenshot 2020-10-30 at 13 37 39

Search
Screenshot 2020-10-30 at 14 44 45

### DIFF
--- a/components/Feedback/ArticleItem.js
+++ b/components/Feedback/ArticleItem.js
@@ -2,8 +2,7 @@ import React from 'react'
 import { css } from 'glamor'
 
 import withT from '../../lib/withT'
-
-import { Interaction, colors } from '@project-r/styleguide'
+import { Interaction, useColorContext } from '@project-r/styleguide'
 
 const { P } = Interaction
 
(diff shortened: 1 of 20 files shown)


components/Search/DocumentResult.js
```diff
@@ -30,13 +22,25 @@ export default withT(({ t, node }) => {
     <ActionBar mode='feed' document={node.entity} />
   ) : null
 
+  const highlightStyle = useMemo(
+    () =>
+      css({
+        '& em': {
+          background: colorScheme.getCSSColor('overlayInverted'),
+          color: colorScheme.getCSSColor('textInverted'),
```
tpreusse, line 30: Align with highlights in comments: 
```suggestion
          background: colorScheme.getCSSColor('alert'),
          color: colorScheme.getCSSColor('text'),
```
or change comments to align: https://github.com/orbiting/styleguide/pull/335/files#diff-e99c1fab5175405325c66e78406abfd011c4dbb0522ec4aa6f553b39b30a3ecdR125-R126

tpreusse (approved):
Generally looks good. A few consistency notes, needs testing on staging.

components/Search/DocumentResult.js
ovbm: The problem was that alert is not visible enough on dark mode. I'm going to change the alert colors a bit for them to be usable both as highlight Blocks as well as highlight text.

components/Feedback/ArticleItem.js
```diff
@@ -18,39 +17,37 @@ const styles = {
...
     right: 0
   })
 }
 
-export const NoResultsItem = ({ title }) => (
-  <P style={{ color: colors.disabled }}>{title}</P>
-)
+export const NoResultsItem = ({ title }) => {
+  const [colorScheme] = useColorContext()
+  return <P {...colorScheme.set('color', 'disabled')}>{title}</P>
```
tpreusse, line 27: Here an additional span is necessary to ensure that color disabled wins:
```suggestion
  return <P><span {...colorScheme.set('color', 'disabled')}>{title}</span></P>
```

Because [`Interaction.P`](https://github.com/orbiting/styleguide/blob/26da62022f3bb887e1b514ad07258cc1d0d97553/src/components/Typography/Interaction.js#L105-L117) does `{...colorScheme.set('color', 'text')}` are clear winner can't be determined and now depends on which css rules get injected last, which is not explicitly defined. It's not a matter of prop order but when the css side effect happens—including during SSR and css text transfer from first SSR render.

PR #504, opened 2020-11-09
Custom Header Color Scheme Everywhere

tpreusse:
Allow the user to always pick the header color scheme and only override content color scheme when necessary—not yet supported or page is explicitly set to dark mode by editors.

Screenshot 2020-11-09 at 15 30 43
Screenshot 2020-11-09 at 15 29 51


ovbm (approved):
!important XD oh well...

PR #514, opened 2020-11-20
Add back "automatic option" except for app

ovbm:


components/Frame/DarkmodeSwitch.js
```diff
@@ -67,18 +71,20 @@ const DarkmodeSwitch = ({ colorSchemeKey: pageColorSchemeKey, t }) => {
...
-              }}
-            >
-              {t('darkmode/switch/auto')}
-            </Radio>*/}
+            {!inNativeApp ? (
+              <Radio
+                checked={!colorSchemeKey}
+                onChange={() => {
+                  // default is undefined
+                  setColorSchemeKey()
```
tpreusse, line 82:
```suggestion
                  setColorSchemeKey('auto')
```

components/Frame/DarkmodeSwitch.js
```diff
@@ -67,18 +71,20 @@ const DarkmodeSwitch = ({ colorSchemeKey: pageColorSchemeKey, t }) => {
...
-              onChange={() => {
-                // default is undefined
-                setColorSchemeKey()
-              }}
-            >
-              {t('darkmode/switch/auto')}
-            </Radio>*/}
+            {!inNativeApp ? (
+              <Radio
+                checked={!colorSchemeKey}
```
tpreusse, line 79:
```suggestion
                value='auto'
                checked={!colorSchemeKey}
```

components/Frame/DarkmodeSwitch.js
```diff
@@ -67,18 +71,20 @@ const DarkmodeSwitch = ({ colorSchemeKey: pageColorSchemeKey, t }) => {
...
+                onChange={() => {
+                  // default is undefined
```
tpreusse, line 81:
```suggestion
                  // ToDo activating auto by default
                  // - handle all «ToDo activating auto» comments
                  // - rm explicit auto value
```

components/Frame/DarkmodeSwitch.js
ovbm: shouldn't it still be default white, with just the option to set it as auto?

tpreusse: Well here we need to explicitly save `auto` into the local storage.

When there is no local storage—or a empty value—the frontend is configured to activate light (via color synch & head script)

ovbm: ah logo...

tpreusse (approved):

PR #515, opened 2020-11-20
App V2 Integration

tpreusse:
New light weight messages API for next app version:
- `MessageSync`
  - send `isSignedIn` boolean instead of whole me state
  - receive `onPushRegistered` with token from native app


components/Article/ActionBarOverlay.js
```diff
@@ -7,21 +7,12 @@ const ACTIONBAR_FADE_AREA = 400
 const ActionBarOverlay = ({ children, audioPlayerVisible, inNativeApp }) => {
   const [colorScheme] = useColorContext()
   const [overlayVisible, setOverlayVisible] = useState(false)
-  const [bottomPosition, setBottomPosition] = useState()
   const lastY = useRef()
   const diff = useRef(0)
 
-  useEffect(() => {
```
ovbm: I think we still need this for mobile browsers (non-native-app).

tpreusse: in browser it's always 44 anyhow – doesn't make a difference

ovbm: true!

PR #521, opened 2020-12-04
Removes linkRule from Dialoge title and a couple of other places

ovbm:
linkRule is deprecated (it has a hover with secondary color, which is deprecated), should be removed everywhere and replaced by the new <A> element that is exported from styleguide.

### DIFF
--- a/components/Bookmarks/Page.js
+++ b/components/Bookmarks/Page.js
@@ -13,7 +13,7 @@ import {
   fontStyles,
   Center,
   Interaction,
-  linkRule,
+  A,
   plainButtonRule
 } from '@project-r/styleguide'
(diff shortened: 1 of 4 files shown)


annatraussnig (approved):

components/Bookmarks/Page.js
```diff
@@ -108,9 +108,7 @@ const Page = ({ t, me }) => {
                 {
                   feedLink: (
                     <Link route='feed' key='link'>
```
tpreusse, line 110: Needs `passHref` when it's not a native a tag:
https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag

PR #526, opened 2020-12-15
Query and propagate published, adminUnpublished

patrickvenetz:
This Pull Request will include updated styleguide and some additional translations.

Depends on https://github.com/orbiting/styleguide/pull/355 and https://github.com/orbiting/backends/pull/547

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
   },
   "dependencies": {
     "@next/bundle-analyzer": "^9.5.1",
-    "@project-r/styleguide": "https://s3.eu-central-1.amazonaws.com/republik-assets-dev/npm/project-r-styleguide-pae-b6870f4.tgz",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -102,7 +102,7 @@
   },
   "dependencies": {
     "@next/bundle-analyzer": "^9.5.1",
-    "@project-r/styleguide": "10.8.0",
+    "@project-r/styleguide": "https://s3.eu-central-1.amazonaws.com/republik-assets-dev/npm/project-r-styleguide-pae-b6870f4.tgz",
```
ovbm, line 105: replace with new version number, then` npm i` and add `package.json` and `package-lock.json`

ovbm (approved):

package.json
patrickvenetz: Done.

PR #539, opened 2021-02-04
adds Logo to SeriesNavButton

ovbm:


components/Article/SeriesNavButton.js
```diff
@@ -59,13 +62,9 @@ const styles = {
       fontSize: 18
     }
   }),
-  arrow: css({
```
tpreusse, line 62: `styles.arrow` ist still used, please remove if no longer needed or re-instate if accidentally deleted.

ovbm: done

components/Article/SeriesNavButton.js
```diff
@@ -90,12 +89,31 @@ const SeriesNavButton = ({ t, series, router }) => {
         }}
       >
         <span {...styles.title}>
+          {series.logo && (
+            <>
+              <img
+                {...styles.logo}
+                src={series.logo}
```
tpreusse, line 96:
```js
import { imageResizeUrl } from 'mdast-react-render/lib/utils'
```

For the case that SVGs are not used: we should resize to 48 height (2x of actual height for retina):
```suggestion
                src={imageResizeUrl(series.logo, 'x48')}
```

The asset server will ignore resize instructions for SVG so we do not need to check here.

components/Article/SeriesNavButton.js
```diff
@@ -90,12 +89,31 @@ const SeriesNavButton = ({ t, series, router }) => {
...
+                {...colorScheme.set(
+                  'display',
+                  series.logoDark ? 'displayLight' : 'block'
+                )}
+              />
+              {series.logoDark && (
+                <img
+                  {...styles.logo}
+                  src={series.logoDark}
```
tpreusse, line 105:
```suggestion
                  src={imageResizeUrl(series.logoDark, 'x48')}
```

tpreusse (approved):
General approval for series nav, one more thing: needs a resize to max 48 height in the case pngs or jpgs are used.

PR #553, opened 2021-02-22
fix(NativeApps): Scroll to top after pushing route in MessageSync

patrickvenetz:
When app pushes a route, scroll to top afterwards.

### DIFF
--- a/components/NativeApp/MessageSync.js
+++ b/components/NativeApp/MessageSync.js
@@ -108,7 +108,9 @@ const MessageSync = ({ upsertDevice, me, client }) => {
       } else if (content.type === 'authorization') {
         checkPendingAppSignIn()
       } else if (content.type === 'push-route') {
-        Router.pushRoute(content.url.replace(PUBLIC_BASE_URL, ''))
+        Router.pushRoute(content.url.replace(PUBLIC_BASE_URL, '')).then(() => {
+          window.scrollTo(0, 0)
+        })
       } else if (content.type === 'osColorScheme') {
         if (content.value) {
           setOSColorScheme(content.value)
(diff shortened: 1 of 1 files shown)


tpreusse (approved):

ovbm:
Did you test if this doesn't break navigating to a specific comment from a notification?

patrickvenetz:
(Testing pending.)

tpreusse:
Scroll to comment should be fine, we already wait 50ms before scrolling to the comment for other reasons:

https://github.com/orbiting/republik-frontend/blob/2e69fc4b1a36b832202f35d91d4212c6b17902bd/components/Discussion/Comments.js#L168-L175

But there is another issue. I'll add a suggestion. Thanks for the reminder of our scroll to cases @ovbm.

components/NativeApp/MessageSync.js
```diff
@@ -108,7 +108,9 @@ const MessageSync = ({ upsertDevice, me, client }) => {
       } else if (content.type === 'authorization') {
         checkPendingAppSignIn()
       } else if (content.type === 'push-route') {
-        Router.pushRoute(content.url.replace(PUBLIC_BASE_URL, ''))
+        Router.pushRoute(content.url.replace(PUBLIC_BASE_URL, '')).then(() => {
+          window.scrollTo(0, 0)
+        })
```
tpreusse, line 113: There are potential `scrollTo` conflicts, to work around this we should back-off when there is a hash in the url:
```suggestion
        const targetUrl = content.url.replace(PUBLIC_BASE_URL, '')
        Router.pushRoute(targetUrl).then(() => {
          if (targetUrl.indexOf('#') === -1) {
            window.scrollTo(0, 0)
          }
        })
```

We should test opening `https://www.republik.ch/konto#newsletter`—used in newsletters—after deploy. Currently works on iOS most of the time.

I also just noticed that next 10 has already improved its scroll logic and seems to do scroll 0,0 by default now. But probably won't get around to do the update until late march (needs routing testing and potentially changes).

ovbm:
Testing results:

Android: Fixes bug, now always on top of feed when navigating back.

iOS: Doesn't introduce regressions. Navigating back preserves previous feed position. Opening notifications with comment reply scrolls to respective comment within the dialogue.

PR #554, opened 2021-03-02
add title-block class

annatraussnig:
add title-block class in article render for style purposes

### DIFF
--- a/components/Article/Page.js
+++ b/components/Article/Page.js
@@ -428,7 +428,7 @@ const ArticlePage = ({
                 <ProgressComponent article={article}>
                   <article style={{ display: 'block' }}>
                     {splitContent.title && (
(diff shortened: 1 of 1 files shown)


tpreusse:
I think the problem with that approach is that you'll target the div that is even around the cover, marked violet:
Screenshot 2021-03-02 at 17 04 22

Instead I think we need to target `TitleBlock` and `Center` element—add a className to those in the styleguide (marked in red).

annatraussnig:
but I think that's the only way to catch the paynote as well – otherwise I need a class on it too.

tpreusse:
Paynotes uses a center too but because of the dark background it might need a special rule, not sure though. Would you set a background or just relative and zIndex?

Issue #94, opened 2018-04-24
Support Umlaut Domains in Profile


tpreusse:
E.g.: öpfel.ch

Currently `valid-url` is used:
https://github.com/orbiting/republik-frontend/blob/ba0d1d6cb3f9126a577fff7a0f78e7e8f4f55048/components/Profile/Contact.js#L5

Switch to a url function from `validator`? Was used to fix the same issue with emails.

maracuja-juice:
Hello Thomas
I would love to try solving some small issues in the republik repos and this seems like a good starting point.
I setup the backend and frontends like described in the docs and was able to login in with a test account. Also I added a Mailchimp API key to my environment.
For testing this issue I need local access to my profile page but when visiting that page I get the error: "index_not_found_exception".
What do I need to setup to be able to correctly access this page? I wasn't able to find something about in the docs.

It would be great if the steps to get a fully working frontend running will be added to the "How to Run" documentation in the future.

tpreusse:
Hello @maracuja-juice

Awesome. Thank you for your interest in contributing and sorry about the setup issues.

We've just recently simplified our architecture and need to update our docs soon.

The `index_not_found_exception` comes from the backends, make sure to use the latest backend code from this week and then run `yarn run pull:elasticsearch` in the backends root directory to resolve it and initialise the elastic search indexes.

I just added it to the [backends readme](https://github.com/orbiting/backends/commit/f29b40268ef473fcf80e9950995fd381cc627fe7).

Alternatively you may [proxy our production backend](https://github.com/orbiting/republik-frontend#proxy-production-backend) to your local machine test some stuff with real content:
```
npm run yaproxy
```

Kind regards Thomas

maracuja-juice:
Thank you Thomas for the answer and the documentation update. I will give it another go next weekend!

maracuja-juice:
Hi Thomas.
I pulled the newest code and tried to run the command `yarn run pull:elasticsearch`.
Unfortunately it returned the exit code 1 with the following error:

```
ResponseError: invalid_alias_name_exception
    at IncomingMessage.<anonymous> (/home/mau/Documents/Dev/Republik/backends/node_modules/@elastic/elasticsearch/lib/Transport.js:296:25)
    at IncomingMessage.emit (events.js:327:22)
    at IncomingMessage.EventEmitter.emit (domain.js:467:12)
    at endReadableNT (internal/streams/readable.js:1327:12)
[...]

PR #561, opened 2021-03-20
support domains with mutated vowels (Umlaute)

panmau:
This PR addresses #94.

I used the "isURL" function from the validator package and removed the valid-url package from the project.

Per default the function doesn't validate that the URL should have a protocol and `ftp://` is included in the default protocol list. As this is not what we want, I added options to the function call for validating that the protocol should be either `http` or `https`.

For an overview on which URLs the package regards as valid/invalid I recommend to look at their tests: https://github.com/validatorjs/validator.js/blob/master/test/validators.js#L333

I think it would be a good idea to run the new validator rules over all currently defined domains in the database, to see if any of them are non-compliant.

_PS: As I wasn't able to do the backend setup locally yet, I used the `npm run yaproxy ` command to test this change._

### DIFF
--- a/components/Profile/Contact.js
+++ b/components/Profile/Contact.js
@@ -1,5 +1,5 @@
 import React, { Fragment } from 'react'
-import { isWebUri } from 'valid-url'
+import { isURL } from 'validator'
 import { compose } from 'react-apollo'
 import { css } from 'glamor'
 import { Dropdown, Label, Interaction, IconButton } from '@project-r/styleguide'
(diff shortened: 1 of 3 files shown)


ovbm:
Thanks a lot for the PR. @patrickvenetz, any ideas why we had two validation packages in the first place. Any red flags from your side for moving forward with this?

tpreusse:
Thank you for the contribution.

And the hint regarding checking all current DB values. Found some interesting stuff like `https://`, `https://-` and `https://0`. Will clean those up and improve the edit form a bit in a commit. I'll reference it here later.

PR #573, opened 2021-04-28
comment embed

annatraussnig:


components/Article/Page.js
```diff
@@ -235,7 +245,10 @@ const ArticlePage = ({
                   })
                 }
               })
-            : undefined
+            : undefined,
+        withCommentData:
+          process.browser && window ? withCommentData : undefined,
```
tpreusse, line 250: if the graphql wrapper has `ssr: false`—which it seems to have via styleguide—you shouldn't need that `process.browser` or window check. Also if you want to check a window (process.browser is webpack specific but fine in next.js code bases) you need to do `typeof window !== 'undefined'` just `window` will already trigger an exception in node.

annatraussnig: will change. the problem also arises in publikator, though.

components/Marketing/MiniFront.js
```diff
@@ -9,14 +9,16 @@ const MiniFront = ({ data: { loading, error, front }, t }) => {
...
-      )}
+      render={() =>
+        front ? (
+          <RenderFront
+            t={t}
+            isEditor={false}
+            front={front}
+            nodes={front.children.nodes}
+          />
+        ) : null
```
tpreusse, line 20: was this an issue that happened often? or more of a dev edge case? maybe also a darn apollo bug, can stay in if convenient in dev otherwise this might mask a real problem of the marketing front not properly loading.

tpreusse (approved):
Looks good. Please `npm run translations` to add the new `styleguide/comment/header/unavailable` key—I've just added it in the google doc.

Let's 🚢 it

components/Marketing/MiniFront.js
annatraussnig: dev edge case... it could theoretically cause problems if there is no marketing front document defined.

PR #575, opened 2021-04-30
Fix Round

tpreusse:
Changes
- allow custom pay notes trial form in documents on iOS
- [disable FLoC](https://twitter.com/natfriedman/status/1387159870667849731)
- nicer spacing on pages with custom pay notes

### Spacing

| before                                                                                                                                                           | after                                                                                                                                                               |
|------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| screencapture-republik-ch-probelesen-2021-04-30-12_12_01 | screencapture-localhost-3010-probelesen-2021-04-30-12_10_43 |

(couldn't find a case where the br where necessary with pages like about and statutes)

### DIFF
--- a/components/Article/PayNote.js
+++ b/components/Article/PayNote.js
@@ -245,6 +245,8 @@ const disableForIOS = note => {
       inNativeIOSApp:
         isPermissibleIOSCta(note.before.cta) &&
         isPermissibleIOSCta(note.after.cta)
+          ? 'any'
+          : false
     }
   }
 }
--- a/components/Article/PayNote.js
+++ b/components/Article/PayNote.js
@@ -235,12 +235,22 @@ const generateKey = (note, index) => {
   return { ...note, key: `custom-${index}` }
 }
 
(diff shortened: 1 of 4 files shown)


components/Article/PayNote.js
```diff
@@ -235,12 +235,24 @@ const generateKey = (note, index) => {
   return { ...note, key: `custom-${index}` }
 }
 
+const isPermissibleIOSCta = cta => !cta || cta === 'trialForm'
```
annatraussnig, line 238: wouldn't `cta !== 'button'` be less convoluted?

annatraussnig (approved):
LGTM.

components/Article/PayNote.js
tpreusse: Would be shorter but generally I prefer declaring what is allowed: no cta and `trialForm`. IMO easier to understand what the policy is.

tpreusse:
Thanks for the review 🙇

PR #578, opened 2021-05-03
Native Next Routing

annatraussnig:
- [x] finish pages/ links
- [x] fix AreaLink to work with href directly
- [x] delete PathLink
- [x] remove next-routes altogether
- [ ] test, test, test


components/Auth/TokenAuthorization.js
```diff
@@ -62,11 +62,15 @@ class TokenAuthorization extends Component {
...
 
   goTo = (type, email, context) => {
+    const { router } = this.props
     if (this.props.goTo) {
       this.props.goTo(type, email, context)
       return
     }
-    Router.replaceRoute('notifications', { type, email, context })
+    router.replace({
+      pathname: '/benachrichtigung',
```
patrickvenetz, line 71:
```suggestion
      pathname: '/mitteilung',
```

There is no route `/benachrichtigun`. It should be `/mitteilung` due to renaming pages/notifications.js to pages/mitteilung.js

pages/[slug]/[[...suffix]].js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import dynamic from 'next/dynamic'
+import { useRouter } from 'next/router'
+
+const Article = dynamic(() => import('../../components/Article/Page'), {
+  ssr: true
```
Comment (tpreusse) [inline pages/[slug]/[[...suffix]].js:6]: We should add a loading state here and for profile too:
```suggestion
  ssr: true,
  loading: () => <Frame><Loader loading /></Frame>
```

```js
import Frame from '../../components/Frame'
import Loader from '../../components/Loader'
```

To test you can delay the import with the following code:
```js
import('../../components/Article/Page').then((mod) => {
  return new Promise(resolve => {
    setTimeout(() => {
      resolve(mod)
    }, 5000)
  })
})
```

In practice this currently rarely happens because next preload the code most of the time. But could happen on slow networks or if the logic changes.

pages/[slug]/[[...suffix]].js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
```
Comment (tpreusse) [inline pages/[slug]/[[...suffix]].js:1]: Could we rename this to `pages/[...path].js`? Would need small adaption in profile page but IMO worth it to also have this at the root level of the page folder.

pages/[slug]/[[...suffix]].js
```diff
@@ -0,0 +1,21 @@
...
+const Profile = dynamic(() => import('../../components/Profile/Page'), {
+  ssr: true
+})
+
+const Page = props => {
+  const router = useRouter()
+  const {
+    query: { slug }
+  } = router
+  const Component = slug.startsWith('~') ? Profile : Article
```
Comment (tpreusse) [inline pages/[slug]/[[...suffix]].js:17]:
```suggestion
  const {
    query: { slug, suffix }
  } = router
  const Component = slug.startsWith('~') && suffix === undefined  ? Profile : Article
```

or if we rename
```suggestion
  const {
    query: { path }
  } = router
  const Component = path.length === 1 && path[0].startsWith('~')  ? Profile : Article
```

components/Discussion/CommentLink.js
```diff
@@ -28,26 +28,23 @@ export const getFocusRoute = (discussion, comment) => {
...
 
 export const getFocusUrl = (discussion, comment) => {
-  const focusRoute = getFocusRoute(discussion, comment)
-  if (focusRoute) {
-    return `${PUBLIC_BASE_URL}${routes
-      .find(r => r.name === focusRoute.route)
-      .getAs(focusRoute.params)}`
+  const focusHref = getFocusHref(discussion, comment)
+  if (focusHref) {
+    const [protocol, hostname] = PUBLIC_BASE_URL.split('://')
```
tpreusse, line 45: love this line 👏  so much faster than the whole url parse

tpreusse (commented):
Awesome stuff & great job 👏 

Didn't find any bugs but I have a few suggestions below re loading states with dynamic imports and simplifying routes further.

pages/format/[slug].js
```diff
@@ -0,0 +1,3 @@
+import Page from '../../components/Article/Page'
+
+export default Page
```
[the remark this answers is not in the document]
Comment (annatraussnig) [reply pages/format/[slug].js]: Indeed, it could.

components/Auth/TokenAuthorization.js
annatraussnig: 😄 🎣 good catch and thanks!

components/Frame/Popover/UserNav.js
```diff
@@ -199,8 +198,10 @@ const UserNav = ({
                       )}
                       <NavLink
                         {...fontStyles.sansSerifLight16}
-                        route='pledge'
-                        params={{ package: 'DONATE' }}
+                        href={{
+                          pathname: '/angebote',
+                          query: { group: 'DONATE' }
```
patrickvenetz, line 203:
```suggestion
                          query: { package: 'DONATE' }
```

There is no group "DONATE", but a package.

tpreusse:
`router.pathname === '/pledge'` logic in `Header.js` needs adaption to the new file route name in order to hide header pledge button on pledge page. Similar logic exists in `ProlongBox.js` that also needs adaption.


Screenshot 2021-05-21 at 09 20 49

tpreusse:
Found one more route thingy: `router.route !== '/bookmarks'` in `BookmarkButton.js`

components/Frame/ProlongBox.js
```diff
@@ -42,13 +42,14 @@ const ProlongBox = ({ t, prolongBeforeDate, membership, router }) => {
   const [colorScheme] = useColorContext()
 
   if (
-    router.pathname === '/pledge' ||
-    router.pathname === '/cancel' ||
-    router.pathname === '/meta' ||
+    router.pathname === '/angebote' ||
+    router.pathname === '/abgang' ||
+    router.asPath === '/meta' ||
```
tpreusse, line 47: Let's remove this line completely, before we had the cockpit meta was a special page used for prolonging (December 18 / January 19). Since meta is now just a regular section we don't need to treat it specially anymore.

tpreusse (approved):
🎉  let's ship it.

components/Frame/ProlongBox.js
annatraussnig: should we keep cockpit in, though?

annatraussnig: `abgang` and `angebote` I understand, but `cockpit` is kinda the odd one out, here. (then we might want to include stuff like `about`, too. No?)

tpreusse: jup cockpit should stay excluded: it has it's own prolong cta when the membership due

tpreusse: the reason the cockpit is excluded is because it can be used in mailings as a landing page for deciding whether one should prolong and has it's own CTA for prolong. Unlike about.

tpreusse: the reason the cockpit is excluded is because it can be used in mailings as a landing page for deciding whether one should prolong and has it's own CTA for prolong. Unlike about.

PR #581, opened 2021-05-04
fix editbutton alignment when titleAlign is centered

ovbm:
before
Bildschirmfoto 2021-05-04 um 17 04 22

after
Screenshot 2021-05-04 at 17 13 31


patrickvenetz (commented):
LGTM

PR #585, opened 2021-05-10
new footer design

ovbm:
feedback welcome:

- the footer has a 'black' background and is wrapped in a "dark" color-context.
- it responds more elegantly to mobile sizes

desktop
Screenshot 2021-05-10 at 14 15 35

mobile
Screenshot 2021-05-10 at 14 01 40

before:

mobile
Screenshot 2021-05-10 at 14 05 46

desktop
Screenshot 2021-05-10 at 14 06 29


annatraussnig:
The distribution/balance of columns is much more harmonious indeed (both desktop + mobile).

2-3 questions regarding the content:

- how did you decide to remove the `Manifest` and the `Gutschein einlösen` from the footer altogether? I feel like the manifest is kind of important.
- ditto `seit 2018`
- what I would lose though is the `Komplizen` link. This feels a bit deprecated by now. We can always re-add it if we do some Komplizen-thing again in the future.

Otherwise, really like that you split the code.

annatraussnig:
Ah, I also had a looks at `FAQ`. It's mostly technicalities – meaning that harmony allowing, it might make more sense among the `meine Republik`crowd.

ovbm:
- `Manifest` ist out because the new `About` page includes the manifest.
- `Geschenk Einfügen` should probably be added back
- `seit 2018` I didn't really like the middle placing but couldn't come up with something else XD, probably going back in..

annatraussnig:
Remains to be seen, how well it would work in pixels, but another common placement for `seit 2018` would be right under `REPUBLIK`.

package.json
```diff
@@ -19,7 +19,7 @@
     "translations": "npm run translations-raw && git add -p lib/translations.json && git checkout -- lib/translations.json",
     "translations-vote": "gsheets --key=1oBlTiPKIon4p5aek5vEXluWMn0TRwrk-JftznGjTmdQ --title=live --pretty --out components/Vote/translations-vote.json",
     "analyze": "ANALYZE=true next build",
-    "sg:link": "./scripts/sg_link.sh",
+    "sg:link": "./scripts/sg_link.sh","preinstall": "npm run sg:unlink",
```
tpreusse, line 22: make sure to rm before merging

components/Footer/index.js
```diff
@@ -0,0 +1,357 @@
...
+          </FooterNavLink>
+
+          <SocialLinks />
+        </div>
+        <hr {...styles.hr} {...colorScheme.set('borderColor', 'divider')} />
+        <div {...styles.middleRow}>
+          <div {...styles.nav}>
+            <ul {...styles.navList}>
+              <li {...colorScheme.set('color', 'disabled')}>
+                {me ? 'Meine Republik' : 'Mitglied werden'}
```
tpreusse, line 162: use `t` :) And IMO always call it `Meine Republik` because it should also contain a login link and not just joining

components/Footer/index.js
```diff
@@ -0,0 +1,357 @@
...
+              </li>
+              {!!me && (
+                <>
+                  <li>
+                    <FooterNavLink href='/konto'>
+                      <a {...navLinkStyle}>
+                        {t(me ? 'footer/me/signedIn' : 'footer/me/signIn')}
```
tpreusse, line 169: `footer/me/signIn` will never happen because of the ``!!me` outside. Let's keep a login link here, can also go to `/anmelden` instead of `/konto`

components/Footer/index.js
```diff
@@ -0,0 +1,357 @@
...
+              {!inNativeIOSApp && (
+                <li>
+                  <FooterNavLink
+                    href={{
+                      pathname: '/angebote',
+                      query: { group: me ? 'GIVE' : undefined }
+                    }}
+                  >
+                    <a {...navLinkStyle}>
+                      {t(me ? 'footer/me/give' : 'footer/offers')}
```
tpreusse, line 189: we should check for `hasActiveMembership` from `withMemberStatus`. One might be logged in but not yet a active member

tpreusse: potentially we could also show a list of options directly like on the imprint
Screenshot 2021-05-27 at 15 51 38

And since we habe space when not logged in we could show the gift/give link always. Not logged in can also buy for someone else.

tpreusse (commented):
comes a long nicely, a few refinement suggestions

components/Footer/index.js
ovbm: didn't do the options as it's mostly a layout PR.

tpreusse:
I think we need a bit of a visuell fine tuning, especially for desktop

new font-sizes:
12px
14px
16px

currently all 17px
with on exception 16px for «seit 2018»

also why 1000px max width and the 80/40px padding? feel a bit clumsy

ovbm:
What do you suggest for font sizes? I tried to create more of a visual clustering with the new font sizes that distinguishes the address from the navigation items and their titles. I quite like it tbh. Let's discuss quickly at 2!

PR #592, opened 2021-05-28
Add open-source page

trm217:
This PR adds an initial version of the open-source page. The open-source page aims to inform users of how open-source projects are used in our products and how we contribute to the open-source community.

As of now the page constists of multiple subheadings where different aspects of how we use and contribute to OSS.
Before building the page the command `npm run fetch-licenses` has to be run, in order to generate a JSON document, which holds the data shown in the table.


trm217:
## Todo
- Write and add texts for each subsection
- Explore alternatives on how to showcase all used packages.

## Showcase of used packages
Regarding the showcase of all used packages an idea would be to show a card grid, with each card showcasing the dependency name, the publisher (with avatar if available) and the package description. By clicking a card the corresponding repository / npm-page should be opened in a new tab. 
The required data can be obtained via the GitHub or NPM api.

PR #594, opened 2021-06-03
implementation of new series nav

ovbm:


components/ActionBar/index.js
```diff
@@ -144,7 +143,10 @@ const ActionBar = ({
   const displayMinutes = readingMinutes % 60
   const displayHours = Math.floor(readingMinutes / 60)
   const forceShortLabel =
-    mode === 'article-overlay' || mode === 'feed' || mode === 'bookmark'
+    mode === 'article-overlay' ||
+    mode === 'feed' ||
+    mode === 'bookmark' ||
+    mode === 'seriesEpisode'
```
tpreusse, line 149: IMO we should decide if `ActionBar` modes are camelCase or minus separated. I'd vote for camelCase and would change `article-overlay` to `articleOverlay`

components/Article/TrialPayNoteMini.js
```diff
@@ -0,0 +1,92 @@
...
+    margin: 0,
+    ...fontStyles.serifTitle32
+  }),
+  lead: css({
+    marginBottom: 0,
+    ...fontStyles.serifRegular18,
+    [mediaQueries.mUp]: {
+      ...fontStyles.serifRegular21
+    }
+  })
```
tpreusse, line 33: pay note always use interaction font faces, ideally we could just use interaction typography elements from the styleguide

PR #599, opened 2021-06-14
WIP fix notification queries bug

annatraussnig:
As reported by @patrickvenetz, queries happen before a user "exists"o on the frontend, leading to bugs on the backend.

### DIFF
--- a/components/Notifications/NotificationFeedMini.js
+++ b/components/Notifications/NotificationFeedMini.js
@@ -14,6 +14,7 @@ import { notificationsMiniQuery } from './enhancers'
 import { timeFormat } from '../../lib/utils/format'
 import withT from '../../lib/withT'
 import Link from 'next/link'
+import withMe from '../../lib/apollo/withMe'
 
 const dateFormat = timeFormat('%d.%m')
 
@@ -116,8 +117,10 @@ const styles = {
 
 export default compose(
   withT,
+  withMe,
   graphql(notificationsMiniQuery, {
     options: {
+      skip: props => !props.me,
       fetchPolicy: 'cache-and-network'
     }
   })
--- a/components/Notifications/NotificationIcon.js
+++ b/components/Notifications/NotificationIcon.js
@@ -7,6 +7,7 @@ import withT from '../../lib/withT'
 
 import { notificationSubscription, withNotificationCount } from './enhancers'
 import { containsUnread } from './index'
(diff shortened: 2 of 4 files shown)


patrickvenetz:
Test locally on different pages and thus far issue seems resolved on /feed, /benachrichtigungen.

It still occures on article, section and format pages. This might be due to follow feature in action bar?

annatraussnig:
aaaaah very schlau!!! I think this might be it

annatraussnig:
I'll investigate + resolve

annatraussnig:
@patrickvenetz regarding articles/section/formats, I think that's those 2 are the culprits: 
- https://github.com/orbiting/republik-frontend/blob/master/components/Article/graphql/getDocument.js#L12-L16
- https://github.com/orbiting/republik-frontend/blob/master/components/Article/graphql/getDocument.js#L31-L35

I see 2 options:
1. split the query and run the subscription bits only if `me` is defined
2. adjust the backend to cope with a missing `me` more gracefully
(Is there a 3rd option I'm forgetting @tpreusse?)

annatraussnig:
actually could it be that only the article query causes issues? If yes we can revert the other changes.

tpreusse:
Generally I think the backends needs to deal with handling queries gracefully. Only for `subscribeToMore` and mutations the frontend should avoid making them unless allowed.

patrickvenetz:
Thank you. 

Both of those query props will swallow unauthenticated queries (returning empty arrays).
- [subscribedBy](https://github.com/orbiting/backends/blob/master/packages/subscriptions/graphql/resolvers/Document.js#L26-L28)
- [unreadNotifications](https://github.com/orbiting/backends/blob/master/packages/subscriptions/graphql/resolvers/Document.js#L76-L78)

These seem not the culprits we're looking for.

patrickvenetz:
[subscription to `notifications` is a more likely candidate] to be the culprit.

annatraussnig:
This shouldn't be called once me is undefined, but i'll check again

annatraussnig:
I'm starting to feel like a ghost buster indeed. And I still can't reproduce the bug. Will ping Pae (again) after his lunch break.

patrickvenetz:
I think we just found our [proton pack](https://en.wikipedia.org/wiki/Proton_pack).

PR #610, opened 2021-08-04
Add a button that allows to resubscribe to the newsletter

trm217:
## Changes
- If the user is unsubscribed a button is shown under the info-text, which allows the user to reactivate the newsletter. Once the button has been clicked either a success-message or an error are shown.
- If the user reloads the page while the status is still pending a different info-text is shown and the button text is changed to "send again".

### Reactivate button when the user is unsubscribed

Screenshot 2021-08-04 at 15 25 43

### After the reactivate button has been clicked

Screenshot 2021-08-04 at 15 37 02

### If the user reloads the page while the subscription status is still pending

Screenshot 2021-08-05 at 11 00 59

### DIFF
--- a/components/Account/NewsletterSubscriptions.js
+++ b/components/Account/NewsletterSubscriptions.js
@@ -15,7 +15,7 @@ import {
   Button
 } from '@project-r/styleguide'
 import { withMembership } from '../Auth/checkRoles'
-import { newsletterFragment, userNewslettersFragment } from './enhancers'
+import { newsletterFragment, newsletterSettingsFragment } from './enhancers'
 
 const NoBox = ({ children, style: { margin } = {} }) => (
   <div style={{ margin }}>{children}</div>
@@ -40,10 +40,10 @@ const styles = {
 export const RESUBSCRIBE_EMAIL = gql`
   mutation resubscribeEmail($userId: ID!) {
     resubscribeEmail(userId: $userId) {
-      id
-      status
+      ...NewsletterSettings
     }
   }
+  ${newsletterSettingsFragment}
 `
 
 export const UPDATE_NEWSLETTER_SUBSCRIPTION = gql`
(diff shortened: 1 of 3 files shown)


components/Account/NewsletterSubscriptions.js
```diff
@@ -80,7 +98,59 @@ const NewsletterSubscriptions = props => (
...
-              <P>{t('account/newsletterSubscriptions/unsubscribed')}</P>
+              <Mutation mutation={RESUBSCRIBE_EMAIL}>
+                {(mutate, { loading, error, data: mutationData }) => (
+                  <>
+                    {status === 'unsubscribed' && (
+                      <P>{t('account/newsletterSubscriptions/unsubscribed')}</P>
+                    )}
+                    {/* Show if the status has been set to pending */}
+                    {!error &&
+                      mutationData?.resubscribeEmail?.status === 'pending' && (
```
tpreusse, line 109: Is the `!error` check here really necessary? I feel like this could be removed to simplify things.

components/Account/NewsletterSubscriptions.js
```diff
@@ -80,7 +98,59 @@ const NewsletterSubscriptions = props => (
...
+                      <P>
+                        {t(
+                          'account/newsletterSubscriptions/resubscribeEmailPending'
+                        )}
+                      </P>
+                    )}
+                    <div style={{ marginTop: 10 }}>
+                      {!mutationData && (
+                        <>
+                          {error && <ErrorMessage error={error} />}
```
tpreusse, line 125: As far as I can see this could be moved up and simplified:
```suggestion
                    {!mutationData && <div style={{ marginTop: 10 }}>
                      {error && <ErrorMessage error={error} />}
```
(also needs adjustment at the end, changes: rm fragment and move condition around div)

tpreusse (commented):
Looking good. A few simplification suggestions. I'll look at the texts after the daily.

components/Account/NewsletterSubscriptions.js
trm217: You're right, one can assume error to be undefined if mutationData is defined.

components/Account/NewsletterSubscriptions.js
```diff
@@ -80,7 +98,52 @@ const NewsletterSubscriptions = props => (
...
+                      </P>
+                    )}
+                    {!mutationData && (
+                      <div style={{ marginTop: 10 }}>
+                        {error && <ErrorMessage error={error} />}
+                        {loading && <InlineSpinner size={40} />}
+                        {!loading && (
+                          <Button
+                            primary
+                            disabled={loading}
```
tpreusse, line 126: Not necessary, the whole button is not rendered when loading::
```suggestion
```

tpreusse (approved):
Nice work.

We also need to run `npm run translations` one more time before merging.

PR #613, opened 2021-08-16
Refetch front if last refetch was before 5am today

ovbm:
I removed the state, seems to be simpler and still equivalent in function. Or am I overlooking smthng?

### DIFF
--- a/components/Front/index.js
+++ b/components/Front/index.js
@@ -118,7 +118,7 @@ export const RenderFront = ({ t, isEditor, front, nodes }) => {
   )
 }
 
-const lastFetchAttemptAtbyPath = {}
+let lastMountAt = undefined
 
 const Front = ({
   data,
@@ -132,7 +132,8 @@ const Front = ({
   serverContext,
   isEditor,
   finite,
-  hasOverviewNav
+  hasOverviewNav,
+  shouldAutoRefetch
 }) => {
   const now = new Date()
   const dailyUpdateTime = new Date(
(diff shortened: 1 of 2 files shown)


components/Front/index.js
```diff
@@ -132,6 +134,28 @@ const Front = ({
...
+  const now = new Date()
+  const dailyUpdateTime = new Date(
+    now.getFullYear(),
+    now.getMonth(),
+    now.getDate(),
+    5
+  )
+  const shouldRefetch =
+    front.meta.path === '/' &&
+    lastFetchAttemptAtbyPath[front.meta.path] < dailyUpdateTime
```
tpreusse, line 147: Would switch it around and pass to `isRefetching` as initial state:
```suggestion
  const now = new Date()
  const dailyUpdateTime = new Date(
    now.getFullYear(),
    now.getMonth(),
    now.getDate(),
    5
  )
  const shouldRefetch =
    front.meta.path === '/' &&
    lastFetchAttemptAtbyPath[front.meta.path] < dailyUpdateTime
  const [isRefetching, setIsRefetching] = useState(shouldRefetch)
```
This way we avoid flashing the old front before switching to the loader. `setIsRefetching(true)` can then be removed from the effect.

ovbm: done

components/Front/index.js
```diff
@@ -132,6 +134,27 @@ const Front = ({
...
 }) => {
+  const now = new Date()
+  const dailyUpdateTime = new Date(
+    now.getFullYear(),
+    now.getMonth(),
+    now.getDate(),
+    5
+  )
+  const shouldRefetch =
+    front.meta.path === '/' &&
```
tpreusse, line 145: This might be undefined during loading and will make boom. While `front?.meta?.path` would be an option this would further complicate things because we'd have to wait for the page to load and path to be defined before we know it. I'd switch to using the same expression as in the query var which is always already defined on load:
```
const path = props.path || cleanAsPath(props.router.asPath)
```

tpreusse (approved):

PR #614, opened 2021-08-17
Dialog Sort Tweak

ovbm:
- Dialog sorts by date the first 24h, after that it sorts by "VOTES", highligting the best comments.
- The order is now stored in the query as a string


tpreusse:
Ideally we could transform the `OrderBy` component in `Discussion/Comments.js` from a button into a (next) link. We already have a reload link there which has all the data loaded to generate the correct url.

ovbm:
good idea, will give it a go

ovbm:
now works with backend auto sort: https://github.com/orbiting/backends/tree/feat-dialog-sorting-auto

fuenkchen (approved):

PR #615, opened 2021-08-17
Add hint, when people write unescaped "*" within a word

ovbm:


ovbm:
tested on .love, werks

PR #619, opened 2021-09-01
add a banner that's shown to all users with an app version < 2.0.0

trm217:
### Changes
- A banner is shown to users of the app that use a version < 2.0.0
- For android users a link to the play-store and direct apk download are shown
- For iOS users a link to the app-store is shown.

### Screenshots

Screenshot 2021-09-01 at 16 29 48
Screenshot 2021-09-01 at 16 29 33
Screenshot 2021-09-01 at 16 30 04
Screenshot 2021-09-01 at 16 30 18


ovbm:
There's an issue rendering the banner when setting the useragent to 1.x.x. It only happens on initial render, after navigation it's fine. My guess is that it's got something to do with the backwards-compatibility code.
Screenshot 2021-09-06 at 16 04 59

components/Frame/index.js
```diff
@@ -23,9 +23,7 @@ import {
 import { withMembership } from '../Auth/checkRoles'
 import withMe from '../../lib/apollo/withMe'
 import withT from '../../lib/withT'
-import withInNativeApp, {
-  inNativeAppBrowserLegacy
-} from '../../lib/withInNativeApp'
+import withInNativeApp, { useInNativeApp } from '../../lib/withInNativeApp'
```
ovbm, line 26: If you use the hook, you might as well remove `withInNativeApp` and just use the hook.

components/Frame/index.js
```diff
@@ -105,6 +103,8 @@ const Frame = ({
   isOnMarketingPage,
   pageColorSchemeKey
 }) => {
+  const { inNativeAppLegacy } = useInNativeApp()
```
ovbm, line 106: import inNativeApp here and remove it from props and from line 182

ovbm (approved):

PR #624, opened 2021-09-07
Fix election candidacy preview

fuenkchen:
* preview of election ballot row is only been displayed after submitting new candidacy or changes to candidacy 
* edit link below preview is changed to a button

not done:
* explanation text and link for biography field concerning markdown - reason: it would not be consistent at the moment, e.g. lead text of the markdown page is focused on discussions, and also we do not have such a hint on profile page. so imho, if we want to promote that you can write MD in your biography we should change it on all relevant places.

### DIFF
--- a/components/Vote/ElectionCandidacy.js
+++ b/components/Vote/ElectionCandidacy.js
@@ -10,11 +10,9 @@ import {
   InlineSpinner,
   Interaction,
   Button,
-  IconButton,
   Label,
   mediaQueries
 } from '@project-r/styleguide'
-import { MarkdownIcon } from '@project-r/styleguide/icons'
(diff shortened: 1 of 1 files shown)


annatraussnig:
Markdown explanation: maybe it wouldn't be such a bad idea to include a `MdInput` field (complete with relevant hints, etc.) in the styleguide at some point and replace all fields that support such a pattern with it. This would also be useful if we ever decide to replace said fields with a regular editor.
It's a bit out of the scope of this PR, but it still find it confusing that some fields support markdown but don't mention it as such (wouldn't have guessed it).

annatraussnig (approved):

PR #626, opened 2021-09-10
fixes bug that broke discussion pages

ovbm:
Problem was that getDiscussionHref was returning undefined for discussion pages. Created a new function that includes both getDiscussioHref and getDiscussionPath, which returns a UrlObject.

### DIFF
--- a/components/Discussion/Comments.js
+++ b/components/Discussion/Comments.js
@@ -17,7 +17,7 @@ import DiscussionPreferences from './DiscussionPreferences'
 import SecondaryActions from './SecondaryActions'
 import ShareOverlay from './ShareOverlay'
 import CommentLink, { getFocusHref, getFocusUrl } from './CommentLink'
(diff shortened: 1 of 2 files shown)


components/Discussion/Comments.js
```diff
@@ -96,8 +96,7 @@ const Comments = props => {
   } = props
 
   const router = useRouter()
-  const discussionHref = getDiscussionHref(discussion)
-
+  const discussionHref = getDiscussionUrlObject(discussion)
```
patrickvenetz, line 99: Since you went through the trouble to rename `getDiscussionHref` to `getDiscussionUrlObject`, rename `discussionHref`, too?

patrickvenetz (approved):
LGTM

components/Discussion/Comments.js
ovbm: done

PR #630, opened 2021-09-14
support share image on marketing page

annatraussnig:


felixmichel (approved):

tpreusse:
Should we move this logic to the backend? @fuenkchen — construct the image url for with share text in a resolver?

PR #631, opened 2021-09-16
remove link style-overwrite

trm217:
## Description

Beneath the "Abonnierte Profile" heading the links to the profiles behave in an unexpected way. When the URL is accessed from an other site, the link styling isn't as expected (see first screenshot). However when the page is reloaded, the links are styled as expected (see second screenshot)

## Screenshots

Screenshot 2021-09-16 at 10 39 25
Profile link is black and underlined after accessing the page from a different URL.
(This is fixed with this PR )

Screenshot 2021-09-16 at 10 39 57
Profile link is green as expected after reloading the page

### DIFF
--- a/components/Notifications/SubscribedAuthors.js
+++ b/components/Notifications/SubscribedAuthors.js
@@ -40,6 +40,9 @@ const styles = {
       marginBottom: 0
     }
   }),
+  authorLink: css({
+    cursor: 'pointer'
+  }),
   checkbox: css({
     display: 'flex',
     flexDirection: 'row',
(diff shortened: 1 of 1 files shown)


components/Notifications/SubscribedAuthors.js
```diff
@@ -148,7 +144,7 @@ const SubscribedAuthors = ({
                 >
                   <div {...styles.author}>
                     <Link href={`/~${author.userDetails.slug}`}>
-                      <A {...styles.userLink}>{author.object.name}</A>
+                      <A {...styles.authorLink}>{author.object.name}</A>
```
tpreusse, line 147: A few inputs here:
- [avoid custom elements](https://styleguide.republik.ch/dev/colors#avoid-custom-elements)—this rule has been formulated in the context of color schema—but applied generally
- Cursor pointer is missing because the href is not passed to the a tag: add `passHref` to `<Link>`
- IMO we should use `Editorial.A`—too many green links are annoying :D

trm217: I will replace the custom css-rule and use the passHref attribute. 👍🏻

PR #636, opened 2021-09-21
Gallery fixes

annatraussnig:
## prioritise shower dark mode images
Screenshot 2021-09-21 at 16 36 01

## show image group captions in gallery
(when no individual caption is set)
Screenshot 2021-09-21 at 16 29 00
Screenshot 2021-09-21 at 16 29 07

### DIFF
--- a/components/Gallery/ArticleGallery.js
+++ b/components/Gallery/ArticleGallery.js
@@ -1,7 +1,6 @@
 import React, { Component, Fragment } from 'react'
 import PropTypes from 'prop-types'
 import Gallery from './Gallery'
-import get from 'lodash/get'
 import { imageSizeInfo } from 'mdast-react-render/lib/utils'
 import { postMessage } from '../../lib/withInNativeApp'
 import { removeQuery } from '../../lib/utils/link'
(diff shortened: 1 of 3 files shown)


components/Gallery/ArticleGallery.js
```diff
@@ -107,6 +122,7 @@ class ArticleGallery extends Component {
   }
 
   static getDerivedStateFromProps(nextProps) {
+    console.log(getGalleryItems(nextProps))
```
tpreusse, line 125: needs clean up :)

components/Gallery/ArticleGallery.js
```diff
@@ -32,7 +42,12 @@ const findFigures = (node, acc = []) => {
 
 const getImageProps = node => {
   const url = get(node, 'children[0].children[0].url', '')
+  const urlDark = get(node, 'children[0].children[2].url')
   const captionMdast = get(node, 'children[1].children', [])
+  const included =
+    !get(node, 'data.excludeFromGallery', false) &&
+    imageSizeInfo(url) &&
+    imageSizeInfo(url).width > MIN_GALLERY_IMG_WIDTH
```
tpreusse, line 50: optional: we could do the following:

```suggestion
  const included =
    !get(node, 'data.excludeFromGallery', false) &&
    imageSizeInfo(url) &&
    imageSizeInfo(url).width > MIN_GALLERY_IMG_WIDTH
  if (!included) {
    return
  }
```

and then just `.filter(Boolean)` instead of `.filter(i => i.included)`

annatraussnig: I thought about that but it felt less clear in term of what the code's doing.

components/Gallery/ArticleGallery.js
annatraussnig: shame on me.

tpreusse (approved):
LGTM

One bonus task if you feel like it:
`ArticleGallery.js` is the only place we use `lodash/get`, since we now have optional chaining `?.` syntax replace it with that and reduce the dependency :)

annatraussnig:
hhhm but we'd still need to make sure that `children` exists indeed... or am I missing something?

annatraussnig:
e.g. in `children[0].children[0].url`

tpreusse:
@annatraussnig  Conditional chaining works with arrays too:
```
[1]?.[0] === 1
```

annatraussnig:
alright. I done did it

tpreusse (approved):

PR #644, opened 2021-09-24
show prolong box in ios

ovbm:
can be merged if ios review of 2.0.2 goes through

### DIFF
--- a/components/Frame/ProlongBox.js
+++ b/components/Frame/ProlongBox.js
@@ -40,13 +40,11 @@ const dayFormat = timeFormat('%d. %B %Y')
 const ProlongBox = ({ t, prolongBeforeDate, membership }) => {
   const [colorScheme] = useColorContext()
   const router = useRouter()
-  const {
-    inNativeAppWithStripeCompatibility,
-    inNativeIOSApp
-  } = useInNativeApp()
+  const { isMinimalNativeAppVersion, inNativeIOSApp } = useInNativeApp()
 
   if (
-    (inNativeIOSApp && !inNativeAppWithStripeCompatibility) ||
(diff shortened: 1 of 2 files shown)


tpreusse:
IMO let's wait two weeks or so until most people have updated. Or to be fully proper: only allow it for 2.0.2+

ovbm:
yeah, I'll change it to only allow it on 2.0.2

ovbm:
@tpreusse ready for review

components/Frame/ProlongBox.js
```diff
@@ -40,10 +40,10 @@ const dayFormat = timeFormat('%d. %B %Y')
 const ProlongBox = ({ t, prolongBeforeDate, membership }) => {
   const [colorScheme] = useColorContext()
   const router = useRouter()
-  const { inNativeIOSApp } = useInNativeApp()
+  const { inNativeAppWithStripeCompatibility } = useInNativeApp()
 
   if (
-    inNativeIOSApp ||
+    inNativeAppWithStripeCompatibility ||
```
tpreusse, line 46:
```suggestion
    (inNativeIOSApp && !inNativeAppWithStripeCompatibility) ||
```

tpreusse (commented):
One logic mistake. Otherwise looks good.

Not sure if we should cache `inNativeAppWithStripeCompatibility` in a `useMemo`. Now it is computed every time `useInNativeApp` is used.

ovbm:
I can also remove it from the hook and compute it directly in the prolong box

ovbm:
Done, tested in safari with setting userAgent to RepublikApp/2.0.2 vs  RepublikApp/2.0.1.

tpreusse (approved):

PR #648, opened 2021-10-01
Rewrite Profile Page

tpreusse:
Move profile to separate page with rewrite.

This seems to work. Maybe we can already deploy it today separately.


trm217 (commented):
Looks good to me and seems to be working as expected. 

I am unsure about the URL `/~/{username}`, I think a URL such as `/users/{username}` would be more descriptive and fitting. Since the rewrite doesn't change the URL in the browser bar, I suppose it wouldn't pose a problem since the visible URL will remain the same. What are your thoughts on this?

PR #657, opened 2021-10-21
adds trial back into app

ovbm:
Adds back trials in the app. 
- Replaces the "angebote" button in the header on marketing page with a anchor link "probe lesen" 
- Replaces all offers with trial forms on the marketing page 
- Removes offer button on articles
- Adds back trial banners in articles inNativeApp
- tweaks marketing page hero component styles a bit

Todo
- [x] Proper tracking 
- [x] Test in native app

marketing page:
Screenshot 2021-10-21 at 09 44 18

press on "Probe lesen" scrolls to
Screenshot 2021-10-21 at 09 44 26

article page:
Screenshot 2021-10-21 at 09 44 42


ovbm:
After reading Apple guidelines I'm pretty confident that we can do this. Apple deleted the following passage from the app store on october 22: 

`Deleted from 3.1.3. “Developers cannot use information obtained within the app to target individual users outside of the app to use purchasing methods other than in-app purchase (such as sending an individual user an email about other purchasing methods after that individual signs up for an account within the app).”`

The one thing we have to make sure is to design the "tiral ended" state very carefully. In that state users shouldn't be able to continue using the app, and we can't add a link for them to purchase stuff. This should be fine, because we are now allowed to send them the offer-links via email.

See netflix as an example for trial ended state:
E-PWrEyUYAQ-GHJ

Apple has announed that buttons to link to external purchases for reader apps will be allowed in 2022. So in 2022, the only thing we'll change is add a button to our "trial ended"-state.

ovbm:
Latest [commit](https://github.com/orbiting/republik-frontend/pull/657/commits/03f84d8e61fb17ead25842bf2753dd38047c3eeb) restores the SignInPage for iOS for versions below 2.1.0. I can't release a new android version due to SDK requirements so if this PR is merged, it will automatically take effect on Android but have no apparent changes on iOS until we release 2.1.0 - 2.1.0 is now ready to test in testflight btw.

pages/index.js
```diff
@@ -31,7 +31,8 @@ const IndexPage = ({ t, isMember, router }) => {
     )
   }
 
-  if (inNativeApp) {
+  // only show marketing in ios app if it's the latest version
+  if (inNativeIOSApp && isMinimalNativeAppVersion('2.1.0')) {
```
tpreusse, line 35:
```suggestion
  if (inNativeIOSApp && !isMinimalNativeAppVersion('2.1.0')) {
```

ovbm: 😅

PR #665, opened 2021-10-28
removes article progress prompt and replaces it with Actionbar callout 

ovbm:
- Removes the previous ProgressPrompt component that rendered a block above the article, wich is an ugly pattern and should be avoided.
- New the Progress Component in the ActionBar handles the request for user consent. 
- If the user hasn't consented to recording the user position, the Progress Component in the ActionBar shows a "Leseposition merken?" (questionmark added after screenshots were taken) Button with the consent buttons:

Screenshot 2021-10-28 at 16 33 00
Screenshot 2021-10-28 at 16 33 04

testing should include:
- [x] test that "Leseposition merken?" Button doesn't appear in the overlay actionbar
- [x] test that after consenting the progress component updates as intended


components/ActionBar/index.js
```diff
@@ -206,24 +206,22 @@ const ActionBar = ({
...
-            noScroll={mode === 'feed'}
-          />
-        ) : (
-          <></>
-        ),
+      element: (
+        <UserProgress
+          documentId={document.id}
+          forceShortLabel={forceShortLabel}
+          userProgress={document.userProgress}
```
annatraussnig, line 213: What was the incentive behind splitting the show/hide logic the component itself and and the `show`attribute?

components/ActionBar/index.js
```diff
@@ -392,14 +390,14 @@ const ActionBar = ({
...
-        ) : null,
-      show: document && document.userProgress && displayMinutes > 1 && !podcast
+      element: (
+        <UserProgress
+          documentId={document.id}
+          userProgress={document.userProgress}
+          displayMinutes={displayMinutes}
+        />
+      ),
+      show: !!document
```
annatraussnig, line 400: Why is the `UserProgress` element coming up twice? (I'm sure there is a good reason, but I don't get it...)

annatraussnig (approved):
I like the new pattern, I tested it some but I don't have the precise constellation of functionalities in mind – so maybe have a last look before deploying.
The code of `UserProgress` is quite hard to follow, with the various logical branches. I'd look into splitting it.

components/ActionBar/index.js
ovbm: the ActionBar component doesnt use the `withProgressApi`, which provides `showConsentPrompt`

components/ActionBar/index.js
ovbm: yes, the first instance is for       `modes: ['articleOverlay', 'feed', 'bookmark', 'seriesEpisode'],`, the second instance is for the second row of the article top. Could be refactored.

PR #669, opened 2021-11-04
quick fix (not perfect), needs proper refactor

ovbm:
before:
Screenshot 2021-11-04 at 11 16 46

after:
Screenshot 2021-11-04 at 11 16 26

It's not perfect, but better - I have another brancht that refactors actionbar - which I can deploy with a more reobust fix later this week.

### DIFF
--- a/components/ActionBar/index.js
+++ b/components/ActionBar/index.js
@@ -313,7 +313,11 @@ const ActionBar = ({
           path: meta.path
         })
       },
-      label: t('PodcastButtons/play'),
+      label:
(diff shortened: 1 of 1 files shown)


tpreusse (approved):
LGTM

One Q: Flex wrap would not be a option here?

ovbm:
It would and I have it in the new branch - Currently the ActionBar has two Rows - the new one reduces that into one with flex-wrap. maybe I can deploy today if the merge goes smoothly.

PR #672, opened 2021-11-08
update bucket count

annatraussnig:
- Increment count in correct tag bucket after publishing a comment
- Fix logic error in how the code gets its count for `All` comments (previous method didn't account for comments without tags).

### DIFF
--- a/components/Discussion/graphql/store.js
+++ b/components/Discussion/graphql/store.js
@@ -34,6 +34,7 @@ export const mergeComment = ({ comment, initialParentId }) => draft => {
   })
 
   bumpCounts({ comment, initialParentId })(draft)
+  bumpTagCounts({ comment, initialParentId })(draft)
 }
 
 // we keep track of which cache keys we've already bumped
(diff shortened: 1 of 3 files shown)


components/Discussion/graphql/store.js
```diff
@@ -84,6 +86,28 @@ export const bumpCounts = ({ comment, initialParentId }) => draft => {
...
+
+  if (!tags?.length) return
+
+  for (const tag of tags) {
+    const bucket = draft.discussion.tagBuckets?.find(b => b.value === tag)
+    if (bucket) {
+      bucket.count++
+    }
+  }
+}
```
tpreusse, line 109: `parentIds` is always present. We also relay on it in `Comments.js`.

```suggestion
  const parentId = comment.parentIds[0]
  const rootComment = parentId
    ? nodes.find(n => n.id === parentId)
    : comment

  const affectedTags = rootComment?.tags

  if (!affectedTags?.length) return

  for (const tag of affectedTags) {
    const bucket = draft.discussion.tagBuckets?.find(b => b.value === tag)
    if (bucket) {
      bucket.count =+ 1
    }
  }
}
```

tpreusse (approved):
LGTM, optional style stuff / too defensive I noticed.

PR #673, opened 2021-11-09
Refactor Document List

tpreusse:
While trying to debug the issue of fetch more in format feeds not working I refactored some code which didn't help but is still nice to have after a careful review & test round.

The workaround I found for fetch more:
https://github.com/orbiting/republik-frontend/commit/468a2acb8524a5675b11721299c99581b5868dd1

It basically ensures that the whole `getFeedDocuments` query of format feeds is re-run and properly subscribed to. The issue also disappeared when `articleRefetch` in `Article/Page` was disabled. My theory is that somehow after a refetch nested queries get broken / no longer receive updates. Remounting the query then seems to help.

While debugging I also explored `typePolicies` for the Query type that merges documents. Following logic worked but did not solve the refetch / data update issue after fetch more.

```js
    new InMemoryCache({
      typePolicies: {
        Query: {
          fields: {
            documents: {
              keyArgs: args => {
                return Object.keys(args).filter(a => a !== 'after')
              },
              merge: (existing, incoming) => {
                // only merge truthy objects
                if (!existing || !incoming) return incoming
                return {
                  ...existing,
                  ...incoming,
                  nodes: [...existing.nodes, ...incoming.nodes].filter(
                    // deduplicating due to off by one in pagination API
                    (node, index, all) =>
                      all.findIndex(n => n.__ref === node.__ref) === index
                  )
                }
              }
            }
          }
        },
```

If we want to get ride of `updateQuery` we also need to handle `search` and probably a bunch of other queries.

### DIFF
--- a/components/Feed/DocumentListContainer.js
+++ b/components/Feed/DocumentListContainer.js
@@ -1,11 +1,10 @@
-import React, { Component } from 'react'
+import React, { useEffect } from 'react'
 import PropTypes from 'prop-types'
-import { Query } from '@apollo/client/react/components'
 import Loader from '../Loader'
 import DocumentList from './DocumentList'
-import noop from 'lodash/noop'
 
 import { documentFragment } from './fragments'
+import { useQuery } from '@apollo/client'
 
 export const documentListQueryFragment = `
   fragment DocumentListConnection on DocumentConnection {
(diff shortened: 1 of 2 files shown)


components/Feed/DocumentListContainer.js
```diff
@@ -63,79 +62,71 @@ export const makeLoadMore = ({
     }
   })
 
-class LifecycleWrapper extends Component {
```
trm217, line 66: Getting rid of class-components: I like it 🔥

components/Feed/DocumentListContainer.js
```diff
@@ -63,79 +62,71 @@ export const makeLoadMore = ({
...
+    getConnection,
+    mergeConnection,
+    mapNodes,
+    placeholder,
+    help,
+    empty,
+    feedProps,
+    refetchOnUnmount,
+    showTotal
+  } = props
```
trm217, line 78:
```suggestion
const DocumentListContainer = ({
    query,
    variables,
    getConnection,
    mergeConnection,
    mapNodes,
    placeholder,
    help,
    empty,
    feedProps,
    refetchOnUnmount,
    showTotal
}) => {
```

If the props object isn't used inside the component, I'd move the deconstruction into the function-parameter

trm217 (approved):
Getting rid of class-components 🔥 I like it!
Everything looks good to me (besides the prop-deconstruction inside the function-body)

tpreusse:
thx for the review @trm217

PR #695, opened 2021-11-30
Urabstimmung fixes

annatraussnig:
- fix labels in gender bar chart
- get members data through component props (query didn't work for logged out users)

### DIFF
--- a/components/Vote/ElectionDiversity.js
+++ b/components/Vote/ElectionDiversity.js
@@ -3,41 +3,8 @@ import { Interaction, Label, fontStyles, Chart } from '@project-r/styleguide'
 import { css } from 'glamor'
 import voteT from './voteT'
 import { deduplicate } from '../../lib/utils/helpers'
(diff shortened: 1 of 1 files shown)


components/Vote/ElectionDiversity.js
```diff
@@ -155,7 +155,11 @@ const MembersGender = voteT(({ members, vt }) => {
             ? { ...item, value: item.value + 1 }
             : item
         ),
-      Object.keys(GENDER).map(key => ({ key, value: 0 }))
+      Object.keys(GENDER).map(key => ({
+        key,
+        value: 0,
+        pos: key === 'weiblich' ? 'left' : 'right'
```
tpreusse, line 161: wouldn't it be more logical to be index based? index `0` left rest right?

annatraussnig: Right you are: there you go.

tpreusse (approved):

PR #700, opened 2021-12-10
Payment / PledgeOption-Selector

ovbm:
navigate to http://localhost:3010/checkout to see different option configurations:
Screenshot 2021-12-10 at 12 51 39

The relevant component is MembershipSelector.tsx. It receives a membershipOptions object-array and renders a button for each option. It doesn't handle Gift memberships (which probably should be displayed similarly to goodies instead of price selection).

@tpreusse 

Open questions also with PROLONG: If someone has payed 360CHF, we should probably show this:
Screenshot 2021-12-10 at 12 57 59

But what if a person has payed 250CHF or 378CHF? What do we show? We shouldn't just show "Eigener Preis (360CHF)" & "Regulär 240CHF" as this would be a down-sell. An option would be to Show just "Eigener Preis (360)CHF" without Regular. Or like this:
Screenshot 2021-12-10 at 13 09 32


ovbm:
Current Problems.

**/angebote?package=PROLONG**
- Reducing "Eigener Preis" deselects all suggestions, selecting "Eigener Preis" throws Typerror: 'selectedSuggestion.price'
- Reason Field is shown even if "Eigener Preis" is higher than 240.-
- Selecting "Gönner" and back to "Eigener Preis" shows InputField Error: "Betrag, mindestens CHF 1240"

**package=ABO&userPrice=1**
- Price is 0 by default (should be 140 or something other than 0)
- Inputting a price higher than 240 switches Layout
- Switching from "Eigener Preis" to "Regular" and back shows InputFieldError: "Beitrag mindestens CHF 1" - Error is not shown on first render.

**/angebote?package=ABO_GIVE adn MONTHLY_GIVE**
- Field selection broken: Can't select more than 1 abo
- Field state broken, can't delete inputfield and add new number

ovbm:
moved to: https://github.com/republik/plattform/pull/10

PR #710, opened 2022-01-04
Statements finishing touches

trm217:


components/Discussion/DiscussionProvider/components/DiscussionMetaHelper.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const { discussion, loading, error } = useDiscussion()
+
+  if (loading || error || !discussion) {
+    return null
+  }
+
+  if (discussion.comments.focus) {
+    const metaFocus = discussion.comments.focus
+
+    console.debug('render comment meta')
```
tpreusse, line 24: rm

components/Discussion/DiscussionProvider/components/DiscussionMetaHelper.tsx
```diff
@@ -0,0 +1,61 @@
...
+          }),
+          description: metaFocus.preview,
+          url: getFocusUrl(discussion, discussion.comments.focus)
+        }}
+      />
+    )
+  }
+
+  if (discussion) {
+    console.debug('render discuission meta')
```
tpreusse, line 43: rm

components/Discussion/DiscussionProvider/components/DiscussionMetaHelper.tsx
```diff
@@ -0,0 +1,61 @@
...
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
+          title: t('discussion/meta/title', {
+            quotedDiscussionTitle: inQuotes(discussion.title)
+          }),
+          url: getFocusUrl(discussion)
+        }}
+      />
+    )
+  }
```
tpreusse, line 56: you don't need to do this because the article page renders and controls the meta when not focussed

components/Discussion/DiscussionProvider/components/DiscussionMetaHelper.tsx
```diff
@@ -0,0 +1,61 @@
...
+ */
+const DiscussionMetaHelper = () => {
+  const { t } = useTranslation()
+  const { discussion, loading, error } = useDiscussion()
+
+  if (loading || error || !discussion) {
+    return null
+  }
+
+  if (discussion.comments.focus) {
```
tpreusse, line 21: should also depend on `router.query.focus` IMO to avoid intermediate state issue

components/StatementsDiscussion/StatementDiscussion.js
```diff
@@ -47,16 +41,9 @@ const StatementDiscussion = ({ t, tagMappings }) => {
...
-        refetch({
-          focusId: undefined
-        })
-      })
-    } else {
-      refetch()
-    }
+    refetch({
+      discussionId: discussion.id
+    })
```
tpreusse, line 46: does this also remove the focus id? IMO we should still do that

components/StatementsDiscussion/StatementNodeWrapper.tsx
```diff
@@ -52,9 +52,20 @@ const StatementNodeWrapper = ({
...
+  const ProfileLink = useMemo(() => {
+    let href
+    // Create a link to the user profile or for anonymous users a link to the comment
+    if (comment?.displayAuthor?.slug) {
+      href = `/~${comment.displayAuthor.slug}`
+    } else {
+      href = getFocusHref(discussion, comment)
+    }
+    return getLinkComponent(href)
+  }, [discussion, comment?.displayAuthor])
```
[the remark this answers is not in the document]
trm217: I suppose we could refactor getFocusHref. When looking at the code of the function I see no reason why the entire discussion is passed to it.

tpreusse, line 67: If you want a quick fix do this:
```suggestion
  const profileHref = comment?.displayAuthor?.slug
    ? `/~${comment.displayAuthor.slug}`
    : getFocusHref(discussion, comment)
  const ProfileLink = useMemo(() => {
    return getLinkComponent(profileHref)
  }, [profileHref])
```
Similar change is possible for focus link. Long term I still don't think this is a good solution and I do like the current API interface with `getFocusHref(discussion, comment)`. Let's discuss this when finishing the refactor and spend some time on context and component API design.

components/StatementsDiscussion/StatementDiscussion.js
trm217: it does not. However I don't think this would be the expected behaviour considering the button is called "reload".
I would like to further go into how we update the url-query based on the user input. However I don't think there is enough time for that now.

components/Discussion/DiscussionProvider/components/DiscussionMetaHelper.tsx
trm217: I don't see how that would help. Since there is a good chance that the comment to focus hasn't been loaded yet, I see no benefit im executing this twice.

trm217: Also as of now we never update the focusId while the page has already loaded.

components/StatementsDiscussion/StatementNodeWrapper.tsx
trm217: Fixed it by passing the hrefs and Link element separatly

components/Discussion/DiscussionProvider/components/DiscussionMetaHelper.tsx
tpreusse: ok, i was more thinking to match the behaviour/check in article page but in the end this probably won't matter. If I find a bug I'll report it again but otherwise fine with leaving as is

components/StatementsDiscussion/StatementDiscussion.js
tpreusse: IMO this is more of a consistency issue: we're doing it this way in the other discussions and the href on the actual a tag is also without focus query. I think the behaviour with removing focus is right and how it has been everywhere in the past – see no reason to change it now.

components/StatementsDiscussion/StatementNodeWrapper.tsx
```diff
@@ -51,10 +43,19 @@ const StatementNodeWrapper = ({
...
+    return format(urlObject)
   }, [discussion, comment])
 
+  const profileHref = useMemo(
+    () =>
+      comment?.displayAuthor?.slug
+        ? `~${comment.displayAuthor.slug}`
+        : focusHref,
+    [comment?.displayAuthor?.slug, focusHref]
+  )
```
tpreusse, line 57: For later: no link would be better than `focusHref` when someone is anonymous (not possible tomorrow).

tpreusse (approved):

components/StatementsDiscussion/StatementNodeWrapper.tsx
trm217: noted in Notion Todos

PR #712, opened 2022-01-11
Next 12 & Jest

tpreusse:
Bump to next 12 and migrate test to jest.

Main change is that we no longer disable special next url module in `next.config.js`. This used to cause issues in old browser but no longer seems to be the case.

Rewriting express code to next middle wares and api routes has been postponed for now. Node 16 upgrade postponed to monorepo.


tpreusse:
Tested with IE 11 and iOS 10. Will launch in the afternoon.

PR #714, opened 2022-01-12
Add a workaround for soft-hyphen not displaying

trm217:
Add a workaround for soft-hyphen not displaying on chromium based browsers on the latest version of MacOS


tpreusse (approved):
Top, danke dir 🙇

